PR #9, opened 2020-02-19
set default AMQP user for Celery as 'airone'.

takano32:
Use 'airone' user for default AMQP setting.


takano32:
It works with 'guest' user at localhost AMQP.

Issue #3, opened 2020-01-29
Add configuration for CI


userlocalhost:
## Goal
Being able to confirm test results for each PR

takano32:
Looks good to me for Close this issue.

PR #11, opened 2020-02-20
Amended CI configuration to be able to work test processing

userlocalhost:
## Abstract

This is for #3 which try to run CI for doing tests on CircleCI.

## Changing points

* Amended configuration for CircleCI because of change of default AirOne settings
* Added badge that describes CI test status of master branch


takano32:
LGTM

What works well on this branch is better than anything else.

PR #13, opened 2020-02-21
Fixed a problem that escaped HTML special characters are shown at edit entry page

### LINKED ISSUE #12: HTML special chars encoded when they are specified in array inputs at edit entry page
In the case of an entry that has array attributes which contain HTML Special values as below,
スクリーンショット 2020-02-21 14 58 57

These values are encoded when you open the edit page of it like this.
スクリーンショット 2020-02-21 15 28 45

userlocalhost:
## Abstract
This fixes #12 

## Changes
* Added a JavaScript function `unescapeHtml` that unescape HTML special characters.
* Wrap each texts which are stored at Jinja2 variable by `unescapeHtml` where they are set to input text value.
* Added identifier for each text input elements to apply Bootstrap style-sheet.

## Result of this change
According to an entry that has HTML special characters for each attributes like this.
スクリーンショット 2020-02-21 16 02 25

I confirmed that every attributes are shown correctly at edit page as below.
スクリーンショット 2020-02-21 16 03 06

### DIFF
--- a/static/js/airone_constants.js
+++ b/static/js/airone_constants.js
@@ -48,3 +48,14 @@ var gettext = function(key, lang) {
     return DEFAULT_DICTIONARY[key];
   }
 };
+
+/* This is necessary because the HTML special characters which are set in the variable in Jinja2
+ * is encoded. When you want to use it in the JavaScript processing, you have to wrap it using
+ * this method to show all text properly. */
+function unescapeHtml(safe) {
+  return safe.replace(/&amp;/g, '&')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>')
+    .replace(/&quot;/g, '"')
+    .replace(/&#39;/g, "'");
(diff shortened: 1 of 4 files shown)


hasebe-teruaki:
Entering a backslash doesn't seem to appear on the edit screen.

・show tab
image

・edit tab
image

userlocalhost:
@hasebe-teruaki Thank you very much of finding it out 👍 . This matter reproduces at current implementation (It means this is an old bug). So I will make an issue for it.

takano32:
Is this isn’t able to handle with safe filter?
https://jinja.palletsprojects.com/en/2.11.x/templates/

userlocalhost:
@takano32 
Thank you for telling useful filter.
But, basically, I think the current implementation that expands variable of jinja2 on the javascript code directly is wrong. I should be able to fix it at #14.

Issue #18, opened 2020-03-09
(BUG) We couldn't create/update entry when array_string attribute is set as mandatory


userlocalhost:
When an Entity has attribute of array_string type as mandatory, an entry of this entity cannot create (or update) even though attribute values are set as below. 
<img width="1196" alt="スクリーンショット 2020-03-09 12 54 15" src="https://user-images.githubusercontent.com/469934/76181691-48905b00-6205-11ea-9655-71ed9a12965a.png">

takano32:
see also #13

userlocalhost:
We realized that following bug is remained.

* A mandatory attribute can be updated with blank value by specifying `- NOT SET-` parameter which describes there is no valid entry.

PR #19, opened 2020-03-09
Changed the way to detect element from exact matching to partial matching.

userlocalhost:
## Abstract
This is for #18. This aims to be able to match with what we want to select even though
class parameter's values are changed.
(c.f. https://api.jquery.com/attribute-contains-selector/)

## Change Points
* Changed the way of detecting element to be able to find element which has specified class parameter partially.
* Added an event handler to remove warning which is appeared before at each input elements.

## Operation Verification
I confirmed followings.

* When I don't specify value at each mandatory attributes, warnings are appeared.
* When I set value at each attributes where warnings were appeared, they would be vanished.
* When I push register button after filling in all mandatory attributes, entry could be created and each values are set properly.

スクリーンショット 2020-03-09 15 26 03
スクリーンショット 2020-03-09 15 25 29

### DIFF
--- a/templates/edit_entry/js.html
+++ b/templates/edit_entry/js.html
@@ -76,7 +76,7 @@ var form_validator = function() {
   // check mandatory attrs of referral entry
   $('.attr_referral[mandatory]').each(function() {
     var all_empty = true;
-    $(this).find('input[class="referral_key"],select[class="attr_value"]').each(function() {
+    $(this).find('input[class*="referral_key"],select[class*="attr_value"]').each(function() {
       if($(this).val()) {
         all_empty = false;
       }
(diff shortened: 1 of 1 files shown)


templates/edit_entry/js.html
```diff
@@ -281,6 +281,12 @@
   $(this).parent().find('li[attr_id="'+ attr_id +'"]').text(text);
 }
 
+/* This is an evenr handler to claear warning (e.g. User doesn't input at mandatory attribute) */
+var clear_table_danger_handler = function() {
+  console.info("[onix/event_clear_table_danger(00)]");
```
takano32: What is onix?

userlocalhost: Thank you for giving a notification. I left a debug code.

takano32:
LGTM 👍

PR #20, opened 2020-03-10
Fixed a bug that mandatory parameter was able to be updated with empty value by specifying '- NOT SET -' value

userlocalhost:
## Abstract
This is for #18. Previous PR #19 left a bug that a mandatory parameter value can be clear by specifying `- NOT SET -` parameter.

## Change Points

* Fixed Javascript code at create/update entry not to send a request when `- NOT SET -` value is specified at mandatory attribute which types `entry`.
* Fixed Django code at input validation check to return HTTP 400 when `- NOT SET -` value is specified at mandatory attribute which types `entry`.

## Operation Verification
I confirmed followings.

* When I try to update an entry with specifying `- NOT SET -` value at `entry` typed attribute, then I can't do it and an error is revealed.

スクリーンショット 2020-03-10 16 25 59

### DIFF
--- a/entry/tests/test_view.py
+++ b/entry/tests/test_view.py
@@ -2963,7 +2963,7 @@ class ViewTest(AironeViewTest):
 
             referral_key = []
             if attr_type & AttrTypeValue['named']:
-                referral_key = [{'data': 'foo', 'index': 0}]
+                referral_key = [{'data': '', 'index': 0}]
 
             # This checks error response would be returned by sending a request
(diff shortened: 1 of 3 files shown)


takano32:
Some find method doesn’t use CSS selector.
Does it can’t use CSS selector?

not ‘class=klass’ but ‘.klass’ .

userlocalhost:
Thank you for your comment. Good catch!

takano32:
LGTM 👍

PR #24, opened 2020-03-16
[WIP] Multi-Database

takano32:
Only use a MASTER database when the application must use it.

At other times, it reads data from multiple SLAVE databases.

### DIFF
--- a/airone/db_router.py
+++ b/airone/db_router.py
@@ -1,11 +1,14 @@
 import random
+import threading
 
 from django.conf import settings
 
-
 class DBRouter:
     def db_for_read(self, model, **hints):
-        return random.choice(settings.AIRONE['DB_SLAVES'])
+        setting = settings.AIRONE['DB_MASTER']
+        if DBSlaves.enable():
+            setting = random.choice(settings.AIRONE['DB_SLAVES'])
+        return setting
 
     def db_for_write(self, model, **hints):
         return settings.AIRONE['DB_MASTER']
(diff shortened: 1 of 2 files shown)


airone/db_router.py
```diff
@@ -15,3 +20,45 @@ def allow_relation(self, obj1, obj2, **hints):
...
+        enable = False
+        thread_id = threading.get_ident()
+        if DBSlaves.dict[thread_id] == DBType.SLAVES:
+            enable = True
+        if DBSlaves.dict[thread_id] == DBType.MASTER:
+            enable = False
+        return enable
+
+
+def airone_salve_database(func):
```
userlocalhost: typo: salve

takano32:
ref. https://docs.djangoproject.com/en/3.0/topics/db/multi-db/

takano32:
To enable `DATABASE_ROUTERS`.
```
DATABASE_ROUTERS = ['airone.db_router.DBRouter']
```

takano32:
When I asked an engineer who was familiar with Django, it seemed common to use [QuerySet.using](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#using) in Django.

This PR closes because it is better to avoid unusual methods, I think.

airone/db_router.py
takano32: Thanks.

takano32:
add some commits on `feature/multidb`

takano32:
close.

PR #25, opened 2020-04-14
This is a feature to get an AttributeValue in the specific Attribute of Entry with a small number of SQL

userlocalhost:
This is necessary especially for making a some custome view on the premise of fixing data structure. In this case, it's better to get attribute value without authorization check for the perspective of performance.

In usual, a latest AttributeValue in the specific Attribute of Entry would be got by following processing.
```
attr = entry.filter(condition_query).first()
if attr and user.has_permission(attr, ACLType.Readable):
  attrv = attr.get_latest_value()
```

But this small processing sends 5 SQL to Database as below.
https://gist.github.com/userlocalhost/16d89769a54e31537fbaa204aef48e33#file-case1_usual_way

The `get_attrv()` method which is added in this PR sends only 1 query at the expense of authorization check of Attribute.
```
attrv = entry.get_attrv(attrname)
```
https://gist.github.com/userlocalhost/16d89769a54e31537fbaa204aef48e33#file-case2_using_get_attrv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## v2.2.0
+
+### Added
+* Added a method in Entry to get an AttributeValue from Entry object with a small number of SQL
+
 ## v2.1.0
 
 ### Added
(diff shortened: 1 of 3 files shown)


takano32 (approved):
LGTM

PR #26, opened 2020-05-11
Added processing to re-register elasticsearch data for entries that refer to an entry which name is changed

userlocalhost:
This fixes a bug that search result is different from Database's one when the entry's name which is referred from search result entry attribute is changed.

These are changes in this PR.

  - [x] Added processing to update elasticsearch data for entries which refer to updated one
  - [x] Added a test to check search result entry name is same with updated one
  - [x] Added a test of new operation's Job (REGISTER_REFERRALS)
  - [x] Fixed API `[get] /api/v1/job` not to show hidden job
  - [x] Added a test for above fix

### DIFF
--- a/api_v1/job/views.py
+++ b/api_v1/job/views.py
@@ -45,12 +45,12 @@ class JobAPI(APIView):
             }
         }
 
-        query = {
-            'user': user,
-            'created_at__gte': time_threashold,
-        }
+        query = Q(
+            Q(user=user, created_at__gte=time_threashold),
+            ~Q(operation__in=Job.HIDDEN_OPERATIONS)
+        )
         jobs = [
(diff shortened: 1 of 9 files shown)


userlocalhost:
This is an operation verification. 

After changing the name of an entry (`E1`) that is referred by another entry (`E2`), I confirmed the changing E1's entry name is reflected at the advanced search result of E2 as below.

This is a result of advanced search for E2 (tmp01).
スクリーンショット 2020-05-12 18 36 04

This changes name of E2 from "tmp01" to "tmp01-changed".
スクリーンショット 2020-05-12 18 36 45

This describes above changing was reflected at advanced search result.
スクリーンショット 2020-05-12 18 37 04

takano32 (approved):
LGTM

Issue #28, opened 2020-06-30
Attribute authorization settings of EntityAtter for individual would not be inherited to the one of Attribute.


userlocalhost:
## Abstract
Once user create an entry via API (post /api/v1/entry), an Attribute value would be invisible when `Nothing` is set on the general authorization of its EntityAttr. Because there is a bug individual authorization never be inherited.

## Reproduction procedure
1. Set `Nothing` on general authorization setting of a specific Attribute (in this case `attr1` of Entity `EntityTest1`) and also set `Full Controllable` to the user of logged in.
  <img width="1195" alt="スクリーンショット 2020-06-30 10 34 50" src="https://user-images.githubusercontent.com/469934/86072885-9d5e2700-babd-11ea-81c9-4bc1237ee9f8.png">
<img width="1191" alt="スクリーンショット 2020-06-30 10 35 09" src="https://user-images.githubusercontent.com/469934/86073019-e7470d00-babd-11ea-9cbd-29550d186bbc.png">

2. Create an Entry of `EntityTest1` through API as below
```
curl -X POST 'http://localhost:8080/api/v1/entry' \
-H 'Authorization: Token <AccessToken>' \
-H "Content-type: application/json" \
-d '{ "entity": " EntityTest1", "name": "test2", "attrs": {"attr1": "foo"} }'
```

Then, you could see that created entry doesn't have any attribute.
<img width="875" alt="スクリーンショット 2020-06-30 10 48 38" src="https://user-images.githubusercontent.com/469934/86073709-49544200-babf-11ea-8532-b7857fedd2d6.png">

But, actually, there is an Attribute and value is also set correctly. The reason why user couldn't see it is that individual authentication configuration is not inherited. You can confirm it when you switch login user to administrative one.
<img width="1183" alt="スクリーンショット 2020-06-30 10 43 51" src="https://user-images.githubusercontent.com/469934/86073543-ee224f80-babe-11ea-985b-b58da67f65b0.png">
<img width="1190" alt="スクリーンショット 2020-06-30 10 43 34" src="https://user-images.githubusercontent.com/469934/86073551-f2e70380-babe-11ea-8e8f-4549195604cc.png">

## Note
This problem happens only when user create an entry via API. I confirmed it doesn't happen through Web UI and also editing via API.

PR #29, opened 2020-07-01
This fixes a problem of ACL inheritance

userlocalhost:
The root cause is permissions of individual users and groups were not inherited when a new Attribute is created from EntityAttr. Previous implementation only inherit permissions of operation user and groups that user is belonged to. This commit fixes this problem by inherit permissions of all users and groups (and also refactored some inefficient implementation).
(c.f. #28)

### DIFF
--- a/api_v1/tests/test_api.py
+++ b/api_v1/tests/test_api.py
@@ -834,5 +834,9 @@ class APITest(AironeViewTest):
         entry = Entry.objects.get(id=resp.json()['result'])
         attr = entry.attrs.first()
 
+        # Check whether only permitted users and groups has authorization
         self.assertTrue(all([u.has_permission(entry, ACLType.Full) for u in users.values()]))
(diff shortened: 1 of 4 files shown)


api_v1/tests/test_api.py
```diff
@@ -794,3 +795,44 @@ def test_edit_entry_with_same_value(self):
...
+        self.assertEqual(resp.status_code, 200)
+
+        # Check whether proper ACL configuration is set to the created Attribute
+        resp_data = resp.json()
+        self.assertTrue(resp_data['is_created'])
+        entry = Entry.objects.get(id=resp.json()['result'])
+        attr = entry.attrs.first()
+
+        self.assertTrue(all([u.has_permission(entry, ACLType.Full) for u in users.values()]))
+        self.assertTrue(all([u.has_permission(attr, ACLType.Full) for u in users.values()]))
```
takano32, line 839: It is better to make sure that the permissions of other users are not relevant.

takano32:
LGTM

PR #41, opened 2020-10-08
Fixed a problem that date value won't be shown at advanced search result

### LINKED ISSUE #40: Date typed value is invalid in a result of advanced search
## The way of reproduction
1. Create an Entity(E1) that has an Attribute(A1) of date type.
2. Create an Entry(e1) and set an arbitrary value (e.g. 2020-01-01) at A1.
3. Request advanced search with E1 and A1.

Then the value which is set in the procedure (2) is expected to be shown, but it doesn't as below.

### Search result of e1(hoge)
スクリーンショット 2020-10-08 12 03 54

### Registered information of e1(hoge)
スクリーンショット 2020-10-08 12 04 03

userlocalhost:
Close: https://github.com/dmm-com/airone/issues/40

This fixes the problem that a value which is registered in a date typed
attribute won't be shown at the result page of advanced search. That was
because the "make_search_results" method that retrieves data which is
registered in an elasticsearch returns string typed value for date
typed one. But template expects this value will be date type.

This removes "date" filter to align with the type which is returned by
'make_search_results". And declares elasticsearch version that we
confirmed to work well (and trivial errors) in the README.

## Verification Results
I confirmed registered `date` typed value at an Entry will be shown at the result of advanced search as below.
スクリーンショット 2020-10-08 15 54 57
スクリーンショット 2020-10-08 15 55 05

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,12 +14,12 @@ Here is the documentation to setup the development environment of AirOne.
 ## Installation of AirOne
 You have to install Python3.5+ to run AirOne like below (for the case of `ubuntu`).
 ```
-user@hostname:~$ $ sudo apt-get install python3 python3-pip virtualenv
+user@hostname:~$ sudo apt-get install python3 python3-pip virtualenv
 ```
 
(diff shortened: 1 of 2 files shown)


takano32 (approved):
LGTM

PR #43, opened 2020-10-14
Added description to CHANGELOG for fixing UI bugfix about date

userlocalhost:
A problem that date value won't be shown at advanced search result was
fixed in the past (see also: #41). But this is not described in the CHANGELOG.
This  commit wrote it.


takano32 (approved):
LGTM

PR #46, opened 2020-10-27
Fixed problem to get AttributeValue which is associated with deleted EntityAttr

### LINKED ISSUE #45: Wrong attribute values might be returned by Attribute.get_attrv when EntityAttr was deleted
The method of `getattr` of Attribute class is a helper method to get latest AttributeValue of specifying Attribute.
It should not return AttributeValue which is related with deleted Attribute. However it happend just after an EntityAttr was deleted and before viewing target entry (before deleting EntityAttr propaget to entry).

### Reproduction procedure

1. Create an entity (E1) with EntityAttr(A1).
2. Create an entry (e1) of E1 and set value (v1) to A1.
3. Delete A1 by editing E1.
4. Run following script through django shell

```
from entry.models import Entry

e = Entry.objects.get(id=2155631)
e.get_attrv('A1').__class__
```

The last processing try to get value of deleted attribute. So it's expected that 'NoneType' value would be returned. But it returns AttributeValue of v1.

userlocalhost:
Close: #45 

This commit fixes a problem that Entry.get_attrv may return
AttributeValue of Attribute which is associated with deleted EntityAttr.

In the previous implementation, get_attrv returns AttributeValue of
active Attribute. But, once EntityAttr would be deleted (technically
is_active flag is down) by editing an Entity, Attribute instances which
is associated with it wouldn't be deleted immediately. That would be
reflected until a changed Entity's Entry is accessed. And deleting an
Attribute of active EntityAttr is also operationally unexpected.

Therefore, this changed condition to get AttributeValues through
`get_attrv` method to return AttributeValue that refers self entry and
also refers active EntityAttr which is specified by name argument.

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -1467,11 +1467,13 @@ class Entry(ACLBase):
     def get_attrv(self, attr_name):
         """This returns specified attribute's value without permission check. Because
         this prioritizes performance (less frequency of sending query to Database) over
-        authorization safety.Don't to use this before permissoin check of specified attribute.
+        authorization safety.
+
+        CAUTION: Don't use this before permissoin check of specified attribute.
         """
         return AttributeValue.objects.filter(
             is_latest=True,
             parent_attr__name=attr_name,
-            parent_attr__is_active=True,
-            parent_attr__entry=self
+            parent_attr__schema__is_active=True,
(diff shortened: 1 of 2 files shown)


takano32 (approved):
LGTM

PR #48, opened 2020-11-16
Fixed a bug not to change referral values when entity was edited

### LINKED ISSUE #47: Referral Entity of an EntityAttr can't be changed by editing Entity
Because of the change of #44, a bug that referral Entity of an EntityAttr can't be changed in editing Entity was embedded.

That PR was made to prohibit to change type of EntityAttr to keep consistency of AttributeValues which are related to it.
But referral Entities of EntityAttr should enable to be changed for effective data management.
(e.g. A case to add Entry to refer from a specific Attribute because of changing of management policy)

userlocalhost:
Close: #47

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## in development
+
+### Fixed
+
+* Fixed a bug not to change referral values when entity was edited
+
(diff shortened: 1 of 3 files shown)


hinashi (approved):
LGTM

PR #55, opened 2020-12-09
Changed custom_view before editing the entry

hinashi:
- Added response options by return value of custom_view function in before edit entry.
- Added argument of custom_view function in before copy entry, because want to handle receive data.


userlocalhost (approved):
It's pretty good refactoring. Thanks!

PR #61, opened 2020-12-28
Fixed the upper limit when using offset of GET entry API

### LINKED ISSUE #60: GET entry API offset is not working
When offset is used in the GET entry API, the list limit is fixed at 100.
(e.g. offset=90 -> list=90~100)

```
curl 'http://127.0.0.1:8000/api/v1/entry?entity=TestEntity&offset=90' \
-H 'Authorization: Token xxxxx'
```

hinashi:
close https://github.com/dmm-com/airone/issues/60


userlocalhost (approved):
LGTM!! 👍 Thank you very much!

PR #69, opened 2021-01-25
Refactored authentication API handler for HTTP basic authentication

userlocalhost:
In the previous implementation sends query twice to the database for
authenticating users in the AccessTokenURI. But it's enough to send it
at once. And there was no test for authentication using HTTP basic
authentication.

This commit refactores custom authentication implementation to reduce
sending unnecessary db-query and adds test for HTTP basic
authentication.

### DIFF
--- a/api_v1/user/views.py
+++ b/api_v1/user/views.py
@@ -2,6 +2,7 @@ from django.utils.decorators import method_decorator
 from django.views.decorators.csrf import csrf_protect
 from django.contrib.auth.models import User as DjangoUser
 
+from rest_framework import status
 from rest_framework.response import Response
 from rest_framework.authtoken.models import Token
 from rest_framework.authentication import BasicAuthentication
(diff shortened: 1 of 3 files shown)


api_v1/user/views.py
```diff
@@ -18,9 +19,11 @@ class AccessTokenAPI(APIView):
     permission_classes = (IsAuthenticated,)
 
     def get(self, request, format=None):
-        user = DjangoUser.objects.get(id=request.user.id)
+        user = AironeUser.objects.filter(id=request.user.id).first()
+        if not user:
```
hinashi: I don't think that validation is necessary because non-authenticated users are excluded in the permission_classes.
https://github.com/dmm-com/airone/blob/df803330d5f2af179a91ebe40872f26954747de3/api_v1/user/views.py#L19

hinashi (approved):
LGTM!

PR #70, opened 2021-01-26
Added param of entry refferal API to reduce filter execution time

hinashi:
Adding the target entity to the search criteria when filtering entries will reduce processing time.

e.g.) specifying parameters
- before
```
curl -w "%{time_total}" 'http://${AirOneURL}/api/v1/entry/referral?entry=TestEntry' \
-H "Authorization: Token xxxx"
```
- entity param
```
curl -w "%{time_total}" 'http://${AirOneURL}/api/v1/entry/referral?entry=TestEntry&entity=TestEntity' \
-H "Authorization: Token xxxx"
```
- target entity param
```
curl -w "%{time_total}" 'http://${AirOneURL}/api/v1/entry/referral?entry=TestEntry&target_entity=RefEntity' \
-H "Authorization: Token xxxx"
```
- both
```
curl -w "%{time_total}" 'http://${AirOneURL}/api/v1/entry/referral?entry=TestEntry&entity=TestEntity&target_entity=RefEntity' \
-H "Authorization: Token xxxx"
```

### DIFF
--- a/api_v1/entry/views.py
+++ b/api_v1/entry/views.py
@@ -75,41 +75,34 @@ class EntryReferredAPI(APIView):
 
     @airone_profile
     def get(self, request):
+        # set each request parameters to description variables
+        param_entity = request.query_params.get('entity')
         param_entry = request.query_params.get('entry')
+        param_target_entity = request.query_params.get('target_entity')
+        param_quiet = request.query_params.get('quiet')
+
+        # validate input parameter
         if not param_entry:
             return Response({'result': 'Parameter "entry" is mandatory'},
                             status=status.HTTP_400_BAD_REQUEST)
 
-        # if entity param exists, add schema name to reduce filter execution time
-        param_entity = request.query_params.get('entity')
+        # declare query to send DB according to input parameters
+        query = Q(name=param_entry, is_active=True)
         if param_entity:
-            entry_list = Entry.objects.filter(name=param_entry, is_active=True,
-                                              schema__name=param_entity)
-        else:
-            entry_list = Entry.objects.filter(name=param_entry, is_active=True)
+            query &= Q(schema__name=param_entity)
 
         ret_data = []
-
-        # if target_entity param exists, add target entity to reduce filter execution time
-        param_target_entity = request.query_params.get('target_entity')
-        target_entity = None
-        if param_target_entity:
-            target_entity = Entity.objects.filter(name=param_target_entity, is_active=True).first()
-
-        for entry in entry_list:
+        for entry in Entry.objects.filter(query):
             ret_data.append({
                 'id': entry.id,
                 'entity': {'id': entry.schema.id, 'name': entry.schema.name},
                 'referral': [{
                     'id': x.id,
                     'name': x.name,
-                    'entity': {
-                        # because getting the schema of the referral entry is slow
-                        'id': target_entity.id if target_entity else x.schema.id,
-                        'name': target_entity.name if target_entity else x.schema.name
-                    },
+                    'entity': {} if param_quiet else {'id': x.schema.id, 'name': x.schema.name},
                 } for x in entry.get_referred_objects(entity_name=param_target_entity)]
             })
+
         return Response({'result': ret_data}, content_type='application/json; charset=UTF-8')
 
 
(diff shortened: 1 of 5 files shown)


api_v1/entry/views.py
```diff
@@ -80,8 +80,34 @@ def get(self, request):
...
+            entry_list = Entry.objects.filter(name=param_entry, is_active=True,
+                                              schema__name=param_entity)
+        else:
+            entry_list = Entry.objects.filter(name=param_entry, is_active=True)
+
         ret_data = []
-        for entry in Entry.objects.filter(name=param_entry, is_active=True):
+
+        # if target_entity param exists, add target entity to reduce filter execution time
+        param_target_entity = request.query_params.get('target_entity')
```
takano32: This part is a little complex.
It need to some discussion to refactoring, I think.

takano32: Sorry, not this above part, but below part.

hinashi: Thank you for checking!
I was refactored implementation of target entity.
It may still be complicated...

takano32: That's it!
I wanted to discuss about for such a nice code.
https://github.com/dmm-com/airone/pull/70/commits/468666217d55813da4838672743d68ae5525094f

takano32 (approved):
LGTM 👍

api_v1/entry/views.py
takano32: > It may still be complicated...

It may be caused by Response JSON Schema.

Now it's easy to understand, in this case.

takano32: Sorry, it may be caused by not "Database Schema" but "Response JSON Schema", I think.

api_v1/entry/views.py
```diff
@@ -80,18 +80,36 @@ def get(self, request):
             return Response({'result': 'Parameter "entry" is mandatory'},
                             status=status.HTTP_400_BAD_REQUEST)
 
+        # if entity param exists, add schema name to reduce filter execution time
```
userlocalhost: This implementation is redundant. It will be more simple implementation by using `Q` object as below.
(c.f. https://docs.djangoproject.com/en/3.1/topics/db/queries/#complex-lookups-with-q-objects)

api_v1/entry/views.py
```diff
@@ -80,18 +80,36 @@ def get(self, request):
...
                 'id': entry.id,
                 'entity': {'id': entry.schema.id, 'name': entry.schema.name},
                 'referral': [{
                     'id': x.id,
                     'name': x.name,
-                    'entity': {'id': x.schema.id, 'name': x.schema.name},
-                } for x in entry.get_referred_objects()]
+                    'entity': {
+                        # because getting the schema of the referral entry is slow
+                        'id': target_entity.id if target_entity else x.schema.id,
```
userlocalhost: I feel a little bit strange to use `target_entity` variable in this loop because it must be same with `x.schema`.

entry/models.py
```diff
@@ -1001,6 +1001,10 @@ def get_referred_objects(self):
                 Q(referral=self, parent_attrv__is_latest=True)
                 ).values_list('parent_attr__parent_entry', flat=True)
 
+        # if entity_name param exists, add schema name to reduce filter execution time
```
userlocalhost, line 1004: It's better to use `Q` object.
(c.f. https://docs.djangoproject.com/en/3.1/topics/db/queries/#complex-lookups-with-q-objects)

userlocalhost (commented):
Thank you very much for making this great improvement for performance. I made a few requests to fix to be more better implementation.

api_v1/entry/views.py
hinashi: Thank you for checking!
I made it simple by using a Q object.

entry/models.py
hinashi: I made it simple by using a Q object.

api_v1/entry/views.py
hinashi: For x.schema.~, the execution time will be longer.
Added a quiet option when entity is not needed.

userlocalhost (approved):
LGTM

PR #76, opened 2021-02-04
Fixed problem that listing entity is too slow

### LINKED ISSUE #75: Problem list entity page might be slow
When user registers many Entities at AirOne, showing the list entity page might be slow because it try to show whole entity at once.

userlocalhost:
Close: #75 

This commit fixed titled problem by adding following features

* Restrict the number of listing entity at once
* Enable to shift range of listing entities
* Enable to narrow down to show entity

### DIFF
--- a/entity/views.py
+++ b/entity/views.py
@@ -5,6 +5,7 @@ import yaml
 
 from django.http import HttpResponse
 from django.http.response import JsonResponse
+from django.db.models import Q
 
 from .models import Entity
 from .models import EntityAttr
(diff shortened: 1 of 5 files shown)


entity/views.py
```diff
@@ -30,9 +31,35 @@
...
+    query = Q(is_active=True)
+    if param_keyword:
+        query &= Q(name__icontains=param_keyword)
+
+    overall_entities = Entity.objects.filter(query).order_by('name')
+    return_entities = overall_entities[index_start:index_start + CONFIG.MAX_LIST_ENTITIES]
+
     context = {
-        'entities': [x for x in entity_objects if user.has_permission(x, ACLType.Readable)]
+        'entities': return_entities,
```
hinashi, line 56: After commit, unauthorized entities are also displayed.

hinashi (changes requested):
Thank you for your wonderful request!
I made request to fix entity's permission problem.

entity/views.py
userlocalhost: Thank you for your minute review.

I started to think it's better to show the entity's name even though it's prohibited to access for the sake of effectiveness of data management.

I considered the pros/cons of comparison with hiding entity name and showing one in the list, there is less problem to expose prohibited entity name because the access for its context is secured, I think. In opposite, there is much problem when it's concealed. For example, decentralizing and multiplexing problem might be happened by creating same (or similar) Entities by other users). And it's easy to grasp overall information which is registered in the AirOne.

So it's make sense to change spec to be able to show whole entities in the list regardless of their permissions, I think.

hinashi (approved):
LGTM!

takano32 (approved):
LGTM

PR #82, opened 2021-02-10
register user with email address

takano32:
register_user.sh doesn't register user email.
this commit makes it registering user with email, `${USERNAME}@example.com`.


userlocalhost (approved):
LGTM

PR #84, opened 2021-02-13
Implement password reset based on Django embedded features

syucream:
Currently Airone users cannot password reset themselves without admin helps. This patch allows them to reset their password with sending email, a typical way.

### login

image

### reset

image

### reset done

image

After checking this view, I got a reset message (typically via email, in this case just via stdout)

### reset confirm

image

### reset complete

image

## reset behaviros

- It automatically expires reset token when I did login or reset password.

## Email settings for testing

You can have email settings in `settings.py`. If you hope to output to just stdout, you can use console backend.

```py
EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
```
https://docs.djangoproject.com/en/3.1/topics/email/#console-backend

Or of course you can use an actual SMTP backend. (in this case, I tried Gmail as a SMTP server)

```py
EMAIL_BACKEND = 'django.core.mail.backends.smtp.EmailBackend'
EMAIL_HOST = 'smtp.gmail.com'
EMAIL_PORT = 25
EMAIL_HOST_USER = 'xxx'
EMAIL_HOST_PASSWORD = 'xxx'
EMAIL_USE_TLS = True
```
https://docs.djangoproject.com/en/3.1/topics/email/#smtp-backend


airone/settings.py
```diff
@@ -266,3 +266,8 @@
 # If log dir is not exists create it.
 if not os.path.exists(os.path.dirname(LOGGING['handlers']['file']['filename'])):
     os.makedirs(os.path.dirname(LOGGING['handlers']['file']['filename']))
+
+# Email settings
+# https://docs.djangoproject.com/en/1.11/topics/email/
+# FIXME it MUST be replaced with a real settings for production
+EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
```
syucream: Any good settings for production usage?

user/forms.py
```diff
@@ -0,0 +1,86 @@
...
+from django.contrib.auth.forms import UserModel, UsernameField
+from django.contrib.auth.tokens import default_token_generator
+from django.contrib.sites.shortcuts import get_current_site
+from django.core.mail import EmailMultiAlternatives
+from django.template import loader
+from django.utils.encoding import force_bytes
+from django.utils.http import urlsafe_base64_encode
+
+
+class UsernameBasedPasswordResetForm(forms.Form):
```
syucream, line 18: It partially imports the original implementation in `django.contrib.auth.forms.PasswordResetForm` so it probably has licensing issue. How should we handle such a situation?

userlocalhost (commented):
Thank you very much for making this feature. Would you please add the description of this feature to the CHANGELOG, and description how to use this feature (e.g. configuration to be able to send e-mails) to the README?

userlocalhost (changes requested):
Could you please show warning message before sending an E-Mail for users who are `user.authenticate_type != AUTH_TYPE_LOCAL`?

This feature send an email for password reset for all kind of users. But it should not send its email and show following warning message for users who don't authenticate with Django's authentication system.

```
This user is configured to authenticate with LDAP. Please retry using that authenticate information.
```

Those users authenticate with LDAP (or other system, we plan to do with OAuth 2.0 in near future). Therefore, we shouldn't send a password reset email for those user whose authenticate type of its User instance is not AUTH_TYPE LOCAL.

hinashi (approved):
LGTM!
I have verified that password reset works in my local env.
Password validation has also been added thanks to django's auth library.
Since there is no validation when creating user and changing  password, we will consider adding it in the future.

userlocalhost (approved):
It's great work. Thanks!

PR #86, opened 2021-02-24
use gunicorn

takano32:
Use gunicorn

* install gunicorn by requirements.txt and requirements-dev.txt
* serve static files with django.contrib.staticfiles.urls

```
$ gunicorn airone.wsgi:application
[2021-02-24 17:42:16 +0900] [24169] [INFO] Starting gunicorn 20.0.4
[2021-02-24 17:42:16 +0900] [24169] [INFO] Listening at: http://127.0.0.1:8000 (24169)
[2021-02-24 17:42:16 +0900] [24169] [INFO] Using worker: sync
[2021-02-24 17:42:16 +0900] [24172] [INFO] Booting worker with pid: 24172
```


hinashi (approved):
Thank you for your investigation.
I have confirmed that static files are delivered even when not in debug mode.
LGTM!

userlocalhost (commented):
Thank you for making this PR. Is it all right to delete empty static files (e.g. static/css/images/ui-icons_444444_256x240.png)?

userlocalhost:
> Thank you for making this PR. Is it all right to delete empty static files (e.g. static/css/images/ui-icons_444444_256x240.png)?

I understood why this PR adds those empty static files. These are referred by jquery-ui (We could easy to realize that by `$ git grep 'ui-icons_444444'` command). And when we run `correctstatic` of Django without adding those file, it would be failed.
And I also realized It's my fail, because when I add jquery-ui library to this repository. I forgot to set them at that time.

We could get those files from [here](https://github.com/jquery/jquery-ui/tree/main/themes/base/images). I'll added after merging this PR.

Thanks.

userlocalhost (approved):

PR #89, opened 2021-03-06
Support '^' and '$' on some search features

syucream:
Support `^`, `$` meta characters on some search features to filter a lot of entries with certain entry names.


syucream:
I will retry it on another PR 🙇

PR #91, opened 2021-03-10
Added the function to make a query request to the slave DB

hinashi:
Currently, AirOne is requesting queries from MasterDB for both writes and reads.
The load on MasterDB becomes a bottleneck when scaling.

The slave DB has a replication delay compared to the master DB.
There is a problem if change all read query requests to slaves.

To solve this, we added a function to change some DB requests to slaves.

- target
```
Entry.get_referred_objects
User.has_permission
```

### DIFF
--- a/airone/tests/test_db.py
+++ b/airone/tests/test_db.py
@@ -1,0 +1,21 @@
+import unittest
+
+from airone.lib.db import get_slave_db
+from django.conf import settings
+
+
+class AirOneDBTest(unittest.TestCase):
+
+    def setUp(self):
+        # this saves original configurations to be able to retrieve them
+        self.orig_conf_db_slaves = settings.AIRONE['DB_SLAVES']
+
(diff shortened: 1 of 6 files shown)


airone/tests/test_db.py
```diff
@@ -0,0 +1,21 @@
...
+
+        # this enables do profiling
+        settings.AIRONE['DB_SLAVES'] = ['slave1', 'slave2']
+
+    def tearDown(self):
+        # this retrieves original configurations
+        settings.AIRONE['ENABLE_PROFILE'] = self.orig_conf_db_slaves
+
+    def test_get_slave_db(self):
+        self.assertIn(get_slave_db(), settings.AIRONE['DB_SLAVES'])
```
userlocalhost, line 21: 👍

userlocalhost (approved):
LGTM

PR #93, opened 2021-03-16
Update Django version to the current LTS version (v2.2.x)

userlocalhost:
This changes Django version from v1.11 to the latest v2.2 LTS. According to this change, this also change MySQL driver from PyMySQL to mysqlclient which is officially supported by Django from v2.2.
(c.f. https://www.djangoproject.com/download/)


takano32 (changes requested):
License change required.

https://github.com/dmm-com/airone/commit/418b7378e6e1caa83fe91c754fe991173c5cc696#diff-c693279643b8cd5d248172d9c22cb7cf4ed163a3c98c8a3f69c2717edd3eacb7

takano32 (approved):
LGTM

PR #97, opened 2021-03-20
Improve attributes search

syucream:
Cont'd from https://github.com/dmm-com/airone/pull/89, I would like to support some rich search options with `^` and/or `$` on attributes search.

# What it works

If you have 2 similar attributes, like this

image

The attribute hint `bc` gets both attributes:

image

In such a case, we can use `^bc` or `bc$` to filter the results more.

image

image

# How it works

Elasticsearch doesn't anchor operators, actually these are handled as a normal character.
https://www.elastic.co/guide/en/elasticsearch/reference/current/regexp-syntax.html#regexp-unsupported-operators

So I changed making ES query to attach `?` to `^` and `$`, then filtering query result with regex on Django app side.


airone/lib/elasticsearch.py
```diff
@@ -211,9 +211,14 @@ def _get_regex_pattern(keyword):
         str: Regular expression pattern of argument
 
     """
+    escaped = prepend_escape_character(CONFIG.ESCAPE_CHARACTERS, keyword)
+
+    # Elasticsearch doesn't support anchor operators,
+    # so it supports both ignoring it and handle it as a normal character
+    chars = escaped.replace('^', '^?').replace('$', '$?')
```
syucream, line 218: Its a breaking change! It makes ES query more ambiguous. If AirOne users use `^` and/or `$` on entry/attribute names and they hope to get exact search result with using these characters, they cannot get the same result.

hinashi (changes requested):
The number of search results remained the number of hits in Elasticsearch.
If below, it is good if the search result is 1.
image

syucream:
I modified it on https://github.com/dmm-com/airone/pull/97/commits/b11488368dad7d0727a5dd65c9a4a40d82b040d2 to count the number of actual entries. It seems to work fine.
image

But, unfortunately it breaks existing behavior on `make_search_results()`. Actually some test cases failed:
```
FAIL: test_search_entries (entry.tests.test_model.ModelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/circleci/project/entry/tests/test_model.py", line 1547, in test_search_entries
    self.assertEqual(ret['ret_count'], 11)
AssertionError: 5 != 11

======================================================================
FAIL: test_search_result_count (entry.tests.test_model.ModelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/circleci/project/entry/tests/test_model.py", line 2159, in test_search_result_count
    self.assertEqual(resp['ret_count'], 10)
AssertionError: 5 != 10

----------------------------------------------------------------------
Ran 459 tests in 189.234s

FAILED (failures=2, skipped=2)
```
https://app.circleci.com/pipelines/github/dmm-com/airone/161/workflows/636f49dd-c63e-4383-856e-346c4aa0afe2/jobs/473

Because `make_search_results()` returns the number of ES search results, NOT the actual number of entries on the page, 😢  in the test case, it expects 11 as a count even if limit = 5 and the number of entries is also 5.

Then how do we handle the number? I think possible options are:
- Change the existing logic to get a count thats the same value with the actual number of entries
  - `make_search_results()` is called by `/api/v1/entry/search`, so it breaks API compatibility.
- Put the problem on ice. We just revert  https://github.com/dmm-com/airone/pull/97/commits/b11488368dad7d0727a5dd65c9a4a40d82b040d2

hinashi:
This is a difficult problem ...
I thought about do _is_matched_entry all the search results of elasticsearch, but it's not smart.

I think it's better to revert b114883. sorry.

We wait for the opinions of others.

syucream:
No problem, you're welcome! Anyway I revert it for now.

hinashi (approved):

userlocalhost (commented):
Thank you very much for your great job! This feature must help many customers, I believe 👍 
But, I'm bother the result that shows wrong hit count.
スクリーンショット 2021-04-29 15 44 05

How about following solution to subtract hit count from the elasticsearch results when user specified special characters (`^` or `$`)?
https://github.com/syucream/airone/pull/5

(I need test to confirm the 'ret_count' that returns make_search_results will be expected when the parameter "hint_attrs" contains the special characters (`^` or `$`) and matches prepared entries).

If you satisfied with above PR, I'm glad for you to merge it.

syucream:
@userlocalhost I fixed it, could you take another look? 🙇

userlocalhost (commented):
Thank you very much for your great work :+1:

I would really appreciate it if you could apply those special search options to the referral attribute values.
(e.g. At the following case, I want to get only entries which matches "^ほげ" at the `ref` attribute, but AirOne returns entries which don't match this filter, that is Entry "hoge" as below).
スクリーンショット 2021-05-06 13 04 54

syucream:
@userlocalhost it caused by attr type limitation that applied only `string` or `text` types, so I mitigated it at https://github.com/dmm-com/airone/pull/97/commits/fc83fdbb466d73749eef5688d2c2063699abfc6b.

What do you think about remaining types, `'date', 'array', 'array_group', 'array_named_object', 'array_object', 'array_string'`? My current implementation just ignores these types.

For more details, about `date` type, I found a wrong behavior (might be a bug?) that seems not to accept attr hints on the current master branch https://github.com/dmm-com/airone/commit/ce3c199f23c595a7c029ee52b57297b3666343e3, like this:
image
then:
image
Whats the correct behavior?

About other `array` types, I don't know which behavior is correct. The current attr hints performs OR matching on multiple elements, like this:
image
then:
image
and also:
image

Should we apply the regex based attr hints like that way, or just ignoring?

userlocalhost (commented):
> What do you think about remaining types, 'date', 'array', 'array_group', 'array_named_object', 'array_object', 'array_string'? My current implementation just ignores these types.

OK, we should make a limitation to be able to apply this feature only for `string` and `text` typed attribute values for now, I think.
But this limitation should be fixed in near future. So, could you please make an issue for this limitation, @syucream?

After making that issue, I'll merge this PR.

Thank you.

syucream:
@userlocalhost Sure! So I revert the latest modification by https://github.com/dmm-com/airone/pull/97/commits/4d32d0b7ec54de41e0e2b1c7aad1316e1c64ba05, and I created a new issue https://github.com/dmm-com/airone/issues/129 to aim a problem on `date` attribute.

Anyway, could you take another look? 🙇

userlocalhost (commented):
LGTM 👍 Thank you very much for your work.

@syucream 
Before merging it. Would you please add description of this feature including restriction (which is described with #129) in the README for next release.

I'll merge it after adding those description will be added in the [CHANGELOG](https://github.com/dmm-com/airone/blob/master/CHANGELOG.md).

PR #98, opened 2021-03-22
Fixed a bug at the background processing of creating Entry

### LINKED ISSUE #95: Failed to create an Entry when same Entry has already been created
A following exception will be happened at the processing of creating an Entry on a background process. 
```
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/user/airone/job/models.py", line 208, in run
    return method(self.id)
  File "/home/user/airone/.venv/lib/python3.6/site-packages/celery/local.py", line 191, in __call__
    return self._get_current_object()(*a, **kw)
  File "/home/user/airone/.venv/lib/python3.6/site-packages/celery/app/task.py", line 375, in __call__
    return self.run(*args, **kwargs)
  File "/home/user/airone/entry/tasks.py", line 115, in create_entry_attrs
    if not any([int(x['id']) == attr.schema.id for x in recv_data['attrs']]):
  File "/home/user/airone/entry/tasks.py", line 115, in <listcomp>
    if not any([int(x['id']) == attr.schema.id for x in recv_data['attrs']]):
AttributeError: 'NoneType' object has no attribute 'schema'
```

It could be reproduced when multiple requests that targets to create same Entry were queued to process at background processes.
The background processing of creating Entry should be safe when that situation was happened.

userlocalhost:
Close: https://github.com/dmm-com/airone/issues/95

The background processing of creating Entry was not idempotent. It
means once its action was run duplicately, an AttributeError exception
would be happened.

The previous its implementation didn't care about that
Entry.add_attribute_from_base might return None when the target entry
has already created or is under processing to be created.
By this commit, the background processing of creating entry will be
idempotent (Once multiple processes run its processing simultaneously or
duplicately, the processing will be done properly and result will be
same).

### DIFF
--- a/entry/tasks.py
+++ b/entry/tasks.py
@@ -114,10 +114,7 @@ def create_entry_attrs(self, job_id):
             # has already been created or is creating by other process. In that case, this job
             # do nothing about that Attribute instance.
             attr = entry.add_attribute_from_base(entity_attr, user)
-            if not attr:
-                continue
-
-            if not any([int(x['id']) == attr.schema.id for x in recv_data['attrs']]):
+            if not attr or not any([int(x['id']) == attr.schema.id for x in recv_data['attrs']]):
                 continue
 
             # When job is canceled during this processing, abort it after deleting the created entry
--- a/entry/tasks.py
+++ b/entry/tasks.py
@@ -109,8 +109,14 @@ def create_entry_attrs(self, job_id):
             if not entity_attr.is_active or not user.has_permission(entity_attr, ACLType.Readable):
                 continue
 
-            # create Attibute object that contains AttributeValues
+            # This creates Attibute object that contains AttributeValues.
+            # But the add_attribute_from_base may return None when target Attribute instance
+            # has already been created or is creating by other process. In that case, this job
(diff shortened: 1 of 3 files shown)


entry/tasks.py
```diff
@@ -109,8 +109,14 @@ def create_entry_attrs(self, job_id):
...
+            # This creates Attibute object that contains AttributeValues.
+            # But the add_attribute_from_base may return None when target Attribute instance
+            # has already been created or is creating by other process. In that case, this job
+            # do nothing about that Attribute instance.
             attr = entry.add_attribute_from_base(entity_attr, user)
+            if not attr:
+                continue
+
             if not any([int(x['id']) == attr.schema.id for x in recv_data['attrs']]):
                 continue
```
userlocalhost: I feel it's a little bit redundant.

hinashi (approved):
LGTM!
I confirmed the reproduction by the following operation.
- Create an entry with GUI
- Next, Edit Attribute an entry with API

userlocalhost:
Thank you for your checking 👍

PR #99, opened 2021-03-22
Fixed not being redirected to the original URL after login

hinashi:
If you visit without authenticated, will be redirected to the login page.
At that time, put the path in 'next' field of query string.
However, there was a problem that the original query string was not included.

Changed to include the query string in 'next' field.

e.g.) `https://AirOneURL/dashboard/search/?query=vmXXXX`

- before
`https://AirOneURL/dashboard/login/?next=/dashboard/search/`

- after
`https://AirOneURL/dashboard/login/?next=/dashboard/search/?query%3DvmXXXX`

PR #101, opened 2021-03-26
Fixed login redirect test failing in python3.5

hinashi:
The following PR caused the python3.5 test to fail.
Because dict doesn't keep the key order if it's less than python 3.6
Changing the order of the query strings does not affect the behavior.
https://github.com/dmm-com/airone/pull/99


userlocalhost (approved):
Thank you very much for fixing!!

PR #102, opened 2021-03-26
Added handler to report celery exception errors

hinashi:
Email settings are required.
If there is no mail setting, it will not be processed.

### DIFF
--- a/airone/celery.py
+++ b/airone/celery.py
@@ -1,5 +1,7 @@
 import os
 from celery import Celery
+from celery.signals import task_failure
+from django.core.mail import mail_admins
 
 # set the default Django settings module for the 'celery' program.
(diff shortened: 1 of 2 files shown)


takano32 (approved):
LGTM

airone/celery.py
```diff
@@ -14,3 +16,24 @@
 
 # Load task modules from all registered Django app configs.
 app.autodiscover_tasks()
+
+
+@task_failure.connect()
```
userlocalhost, line 21: Great!!

userlocalhost (approved):
Thank you for your good work!

It's better to make a test to confirm that an e-mail that has expected context will be dispatched when some exception is happened at celery process. But it's much difficult for now. So I made an issue #115.

PR #104, opened 2021-03-28
Fix getting coverage

syucream:
We got a lower coverage than what we expect:
https://app.circleci.com/pipelines/github/dmm-com/airone/144/workflows/e4d10808-5f5a-4433-a092-4dcbe30f65d7/jobs/423

because the result is generated by `coverage run manage.py test tools.tests`, executed after `coverage run manage.py test`. So it fixes the test order to get correct coverage.
e.g. https://app.circleci.com/pipelines/github/dmm-com/airone/151/workflows/a15c87be-281a-4670-8e53-fa45637c70de/jobs/440

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,8 +15,8 @@ commands =
   pip install -r requirements-dev.txt
   python manage.py makemigrations
   python manage.py migrate
-  coverage run manage.py test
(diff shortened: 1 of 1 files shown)


userlocalhost (approved):
Thank you for your work!

PR #107, opened 2021-04-03
Handle URL encoded params with form-specifc '+'

syucream:
# Background

## Problem

The search form on the nav bar cannot handle whitespaces appropriately. If you input a query with whitespace(s), such as:
image

Then you will get no results with `+` encoded query:
image

This curious behavior sometimes confuse users who hope the search form to eliminate unnecessary whitespaces automatically. In addition it replaces whitespaces with `+` curiously.

## Reason

Its caused by:

- AirOne view doesn't strip unnecessary whitespaces, just throw it into DB query (in that case, `  entry1`) so it returns no results.
- Submitted parameter can contain `+`, not `%20` from whitespaces on a form via web browsers.
- AirOne handles it on JavaScript with `decodeURIComponent()` that doesn't support `+` encoding.

ref. `application/x-www-form-urlencoded`-specific encoding:

- https://stackoverflow.com/questions/2678551/when-to-encode-space-to-plus-or-20
- https://www.ipa.go.jp/security/fy21/reports/tech1-tg/b_09.html

# Resolution

I did:

- Strip whitespaces on Django app side to allow redundant user inputs
- Render the params without `decodeURIComponent()`. Futhermore I removed the JS impl to migrate the complicated small logic to Django side.

After applying the resolution, when a user input ` entry1`:

image

It returns a result with stripped query `entry1`.
image

Unfortunately this patch doesn't remove `+` characters from the URL. I think its reasonable because we don't have any good control point for the behavior (anything with using JS? it should be terrible ...) and the encoding is defined in the spec.


userlocalhost (commented):
We were suffered from this problem for long time. But you gave us a very clever solution that shows how you are competent. Thanks!

hinashi (approved):
LGTM!
I confirmed that the search hits even if there are spaces before and after.

userlocalhost:
I made an issue which is associated with this PR for adding a line this fixing to the CHANGELOG.
https://github.com/dmm-com/airone/issues/109

PR #108, opened 2021-04-03
Advansed search permalink

syucream:
# Overview

The current `/dashboard/advanced_search_result` doesn't accept attr hint parameters so users cannot share filtered search result with something like a permalink.

This PR accepts the parameters, for more details it accepts the parameter, like this:
image

And also it rewrite URL when users filter their search result (currently using `pushState()`

## What I did

- Introduce `attrinfo` a new param on `/dashboard/advanced_search_result`
  - It has the same structure with `hint_attrs` on `Entry.search_entries()`
  - Its encoded in JSON to represent such a complicated structure on GET request parameter
  - It should be replacement for `attr[]`
- Preserve a permalink with `attrinfo` when a user filters a search result with an attr hint
  - Its performed via HTML5 `pushState()`
- Generate a URL with `attrinfo`, not `attr[]` on `/dashboard/advanced_search`

# Discussion points

It minimizes differences to implement it with adding a new GET parameter `attrinfo` in URL-encoded JSON on the search result view, and use HTML5 `pushState()` API to change users address bar to the permalink. In my opinion, it performs a dirty hack on JS code, we need to discuss how to do it more.

- Do we keep to render the search result via Ajax? Is it really needed in JS? In other words, does it require an async process?
- Do we keep to have 2 similar views on `/dashboard/advanced_search_result` and `/api/v1/entry/search`?
- How do we accept attr-hints parameter on `/dashboard/advanced_search_result`? Its a GET method endpoint, and the parameter should be complicated type `[{"name": "xxx", "keyword": "yyy"}, ...]`

### DIFF
--- a/dashboard/tests/test_view.py
+++ b/dashboard/tests/test_view.py
@@ -176,7 +176,12 @@ class ViewTest(AironeViewTest):
         # test to show advanced_search_result page
         resp = self.client.get(reverse('dashboard:advanced_search_result'), {
             'entity[]': [x.id for x in Entity.objects.filter(name__regex='^entity-')],
-            'attr[]': ['attr'],
+            'attr[]': ['attr'],  # an older param will be deprecated
+        })
+        self.assertEqual(resp.status_code, 200)
+        resp = self.client.get(reverse('dashboard:advanced_search_result'), {
+            'entity[]': [x.id for x in Entity.objects.filter(name__regex='^entity-')],
+            'attrinfo': json.dumps([{'name': 'attr'}]),  # A newer param
         })
         self.assertEqual(resp.status_code, 200)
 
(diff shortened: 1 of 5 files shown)


syucream:
I talked with @userlocalhost about this PR and some considerations locally. Rough consensus is:

- We keep using Ajax. It should help to serve filtered advanced_search_result page with a lot of entries to users comfortably
  - For a long term, we can rethink how to realize it, for e.g. switching to React, ...
- We keep 2 endpoints, `/dashboard/advanced_search_result` and `/api/v1/entry/search`. The latter API can be used by AirOne API users; the API is not only for the search result page.
- We don't remove `attr[]` GET request parameter on `/dashboard/advanced_search_result` to keep URL compatibility.
  - If we introduce some high functional parameter like `attrinfo`, existing parameter will be unnecessary. We will plan to deprecate it in the future.

hinashi (approved):
LGTM!
I confirmed the operation in the local env.
Both old and new parameters worked.

userlocalhost (commented):
Thank you very much for dealing with this issue 👍 I have just one request to fix it.

Would you please set filtering keyword at attributes text-boxes after rendering page if those filtering keywords are specified?

Once I opened the page which filters search results by `test` keyword at the `attr` Attribute, the filtering keyword was disappeared as below (User can't grasp which parameters are applied, I think.).
スクリーンショット 2021-04-30 11 34 09

Following is the page before reload.
スクリーンショット 2021-04-30 11 34 01

syucream:
@userlocalhost I fixed it, how about this?

userlocalhost (commented):
Thank you very much for fixing. I found a small glitch that filtering by name won't be applied.

### reproduction method
1. Filter result of advanced search by name, and copy URL
スクリーンショット 2021-06-08 12 25 19

2. Open copied URL, then condition of name filter was vanished
スクリーンショット 2021-06-08 12 28 09

Would you please fix it?

syucream:
@userlocalhost At https://github.com/dmm-com/airone/pull/108/commits/6f9966a648a9da183e8a7303bda11f4b724479bb, the issue should be fixed. Now you can find the results filtered by entry_name, such as:
image

How about this?

userlocalhost (approved):
LGTM 👍 Thank you very much for your great work!

PR #112, opened 2021-04-10
Fix #110 show error messages on create-user

### LINKED ISSUE #110: There is no error handler at the page of User registration
## Background / Problem
When user try to register a new user with empty form. Server responds with HTTP 400 error. But no error message will be revealed because there is no error handling implementation in the processing of JavaScript.

## Requirements of fix
At least, AirOne should implement an error handler to be able to let user to know followings.

* What problem was happened.
* What should user do.

syucream:
https://github.com/dmm-com/airone/issues/110

Show error messages on the form when a user failed to create a new user, like this:
image

### DIFF
--- a/static/js/messagebox.js
+++ b/static/js/messagebox.js
@@ -69,7 +69,7 @@ MessageBox._showMessage = function(text, style) {
     '<button type="button" class="close" data-dismiss="alert" aria-label="close">' +
       '<span aria-hidden="true">&times;</span>' +
     '</button>' + text + '</div>';
-  $(".airone-messagebox").append(content);
+  $(".airone-messagebox").html(content);
   $(".alert").alert();
 };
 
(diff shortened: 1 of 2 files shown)


userlocalhost (changes requested):
Thank you for fixing that problem. Would you please fixing following behavior.

When I submit form without set necessary input form multiply, waning message box which is shown before is remained. I hope old warning message will be clear when submit button is pushed.
スクリーンショット 2021-04-13 16 20 20

syucream:
@userlocalhost it looks an existing problem unrelating to this PR, probably be a bug on JS functions for the MessageBox. I fixed it with https://github.com/dmm-com/airone/pull/112/commits/fbc5fa75a598da1228dea661737bf781943d4eb2 , how about that?

userlocalhost (approved):
It's an excellent fixing. Thanks!

PR #113, opened 2021-04-10
Perform client-side validation on users form

syucream:
Relating to https://github.com/dmm-com/airone/issues/110

Perform the minimum client-side validation on user create/edit forms without requesting to Django App. It should help users to know which's invalid:
image


userlocalhost (approved):
Excellent!! I really appreciate for making measures against invalid input both frontend and server-side 👍

hinashi (approved):
LGTM!

PR #114, opened 2021-04-14
Fix #111 paginate a list of entries

### LINKED ISSUE #111: Requirement of pagination feature at list page of Entries
## Summary
We require to implement pagination feature at the list page of Entries in the specified Entity.

## Background
AirOne implements search feature to be able to search and narrow down Entries user wants to find as below.
スクリーンショット 2021-04-08 11 31 05

## Problem
The Entry list page has limitation to be able to list Entries at most 50 ones even through more than 50+ Entries which are matched with keyword have been registered. Because of this, that narrowing down implementation is not enough for UX in the following use-cases.

* User wants to know how many entries are registered at AirOne that has "template-" prefixed name.
  - User can only know 50 Entries (other registered entries are invisible and unnoticeable.

## Requirements to fix
It's necessary to be able to list Entries beyond the limitation of the number to list Entries at once without spoiling performance.

An idea to solve this problem is pagination, I think.

syucream:
# Overview

It fixes https://github.com/dmm-com/airone/issues/111
It shows a pagination navigator to show a large number of entries(currently its limited by `CONFIG.MAX_LIST_ENTRIES`) similar to the entity page, like this:
image

The pagination works on both `index` and `restore`. And also it works for search result, like Google search, in other words it doesn't support filtering to per page on this pagination.

# TODO

- Remove unnecessary JS implementation
- Confirm if some JS based dynamic renderings.


hinashi (changes requested):
When I opened the link on the next page after narrowing down by keyword, the keyword disappeared.
image
image

hinashi (approved):
LGTM!
Thank you for fixing.

templates/frontend/main.js
```diff
@@ -0,0 +1,2 @@
...
...eateElement("li",{"aria-hidden":!0,key:"separator-".concat(i),className:n},r)):o.push(a),o}),[])}(E||g&&P.length<=g?P:function(t){return v+h>=t.length?t:[].concat((0,S.Z)(t.slice(0,v)),[e.createElement(Pt,{"aria-label":f,key:"ellipsis",onClick:function(e){C(!0);var t=e.currentTarget.parentNode.querySelector("a[href],button,[tabindex]");t&&t.focus()}})],(0,S.Z)(t.slice(t.length-h,t.length)))}(P),l.separator,w)))}));const Nt=(0,i.Z)({root:{},ol:{display:"flex",flexWrap:"wrap",alignItems:"center",padding:0,margin:0,listStyle:"none"},li:{},separator:{display:"flex",userSelect:"none",marginLeft:8,marginRight:8}},{name:"MuiBreadcrumbs"})(Rt);function _t(t){return e.createElement(ge,null,e.createElement(Nt,{"aria-label":"breadcrumb"},e.createElement(xt,{color:"textPrimary"},"Top")),e.createElement(xt,null,"まだないよ"))}function Tt(){return e.createElement(d,{container:!0},e.createElement(d,{item:!0,xs:12},e.createElement(p,null,e.createElement(Et,null))),e.createElement(d,{item:!0,xs:2},e.createElement(p,null,e.createElement(xe,null))),e.createElement(d,{item:!0,xs:10},e.createElement(p,null,e.createElement(_t,null))))}t.render(e.createElement(Tt,null),document.getElementById("app"))})()})();
```
userlocalhost, line 2: This seems like a file which is generated by React. Is this expected?

templates/frontend/main.js.LICENSE.txt
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+/** @license React v17.0.2
+ * react.production.min.js
+ *
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
```
userlocalhost, line 49: This file also seems like a file which is generated by React automatically. Is it necessary to manage it in this repository?

userlocalhost (commented):
Thank you very much for implementing this feature smartly. But there seems to be unnecessary files. Would you please check it?

userlocalhost (approved):
LGTM 👍 

I'll do following work as another PR.
* Remove unnecessary files (e.g. templates/frontend/main.js)
* Add changelog

PR #116, opened 2021-04-17
Fix getting coverage again

syucream:
Current codecov result is still something wrong, seems to have a similar issue to  https://github.com/dmm-com/airone/pull/104


hinashi (approved):
LGTM!

PR #117, opened 2021-04-17
Remove unused http_post_form decorator

syucream:
It seems not to be used anywhere. We can use Django native Form class to validate form data effectively.


userlocalhost (approved):
Good catch!! Thank you for your refactoring 👍

PR #118, opened 2021-04-24
Fix #115 Add a test case ensures sending a failure cause of a Celery task to admins

### LINKED ISSUE #115: Add tests to confirm whether celery task will proceed expectedly
I think it's better to make a test to confirm that an e-mail which has expected context will be dispatched when an error is happened in celery process at #102.

The following document is helpful to make a test to confirm context of sending e-mail.
https://docs.djangoproject.com/en/3.2/topics/testing/tools/#topics-testing-email

syucream:
Close: https://github.com/dmm-com/airone/issues/115

### DIFF
--- a/airone/tests/test_celery.py
+++ b/airone/tests/test_celery.py
@@ -1,0 +1,26 @@
+from airone.celery import celery_task_failure_email
+from django.core import mail
+from django.test import TestCase
+
+
+class CeleryTest(TestCase):
+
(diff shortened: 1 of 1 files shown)


hinashi (approved):
LGTM!
I didn't know there are test case for email of django. Thank you!

userlocalhost (approved):
LGTM. Thanks!

PR #122, opened 2021-05-06
Added necessary files for login/logout

userlocalhost:
I forgot to add template files for authentications in Django v2.2 at the #93


hinashi (approved):
LGTM!

Issue #123, opened 2021-05-10
keyword entries searched in attribute were not found


hinashi:
If you are using entry type attributes when editing an entry
If the search target exceeds the upper limit, it may not be displayed in the results.

The upper limit of the search target is set below.
https://github.com/dmm-com/airone/blob/ce3c199f23c595a7c029ee52b57297b3666343e3/entry/settings.py#L5

e.g.) 172.27.19.1 is missing from search results
![image](https://user-images.githubusercontent.com/5561786/117594202-8d50b280-b178-11eb-8796-fc3e1491535c.png)

PR #124, opened 2021-05-10
Fixed to keyword entries searched in attribute were not found

hinashi:
close: https://github.com/dmm-com/airone/issues/123

When searching for entries, the results are sorted by name to include exact matches.
image


userlocalhost (approved):
LGTM 👍 I confirmed to be able to get small IP address Entry by filtering it in the search box of editing entry form.
スクリーンショット 2021-06-07 19 52 25

Issue #125, opened 2021-05-10
Concurrent execution of tasks added in custom view


hinashi:
There is a problem that the tasks added in the custom view are executed at the same time even if the targets are the same.
When adding a task in a custom view, I am using existing method of create job.

Dependency is confirmed when creating a job.
Since the operation number is different, the dependency confirmation will be skipped.
https://github.com/dmm-com/airone/blob/ce3c199f23c595a7c029ee52b57297b3666343e3/job/models.py#L223-L226

PR #126, opened 2021-05-10
Added job creation method for custom view

hinashi:
close: https://github.com/dmm-com/airone/issues/125

Added a method to create a new edit job in Custom View.


userlocalhost (commented):
Thank you very much for your work. It's good, but the `new_customview_edit` is almost same with calling Job._create_new_job() directly. It's better to call Job._create_new_job() method directly from CustomView (or making wrapper method in CustomView) to solve #126, I think.

PR #130, opened 2021-05-17
set python version as 3.6.9 for pyenv

takano32:
The version of python bundled with Ubuntu 18.04 LTS is 3.6.9.


userlocalhost (approved):
LGTM. Thanks!

Issue #132, opened 2021-05-21
Replace ldap3 library to python-ldap


userlocalhost:
# summary
We want to replace ldap3 library to following python-ldap library.
https://pypi.org/project/python-ldap/

PR #134, opened 2021-05-22
Replace ldap3 with python-ldap

syucream:
https://github.com/dmm-com/airone/issues/132

Replace [ldap3](https://pypi.org/project/ldap3/) with [python-ldap](https://pypi.org/project/python-ldap)


README.md
```diff
@@ -23,6 +23,11 @@ And you have to install RabbitMQ for executing heavy processing as background ta
 user@hostname:~$ sudo apt-get install rabbitmq-server memcached mysql-server python-dev libmysqlclient-dev
 ```
 
+You have to install OpenLDAP library if you use LDAP.
+```
+user@hostname:~$ sudo apt-get install libldap2-dev
```
syucream: python-ldap is a Python binding to OpenLDAP library, so its required.

takano32 (approved):

userlocalhost:
I amazed how fast your work is! I'm so appreciated it.

userlocalhost (changes requested):
Thank you very much for implementing this feature. But it won't work properly with the LDAP server of production environment.

This is execution result in the master branch (commit position is 615747ff40301148be0f2299c4a7691a897efc26) with changing `airone/settings.py` that sets LDAP configuration for production server. It passes authentication by `LDAPBackend.is_authenticated()` method.
スクリーンショット 2021-06-07 11 01 16

After mering this PR to that master branch, I recreated virtualenv and run it with the same `airone/settings.py` as below.
スクリーンショット 2021-06-07 10 40 31
スクリーンショット 2021-06-07 11 00 04

As a result, I couldn't pass authenticatoin with its LDAP server.
Could you please fix it?

syucream:
@userlocalhost First of all, did you install `libldap2-dev` and `libsasl2-dev`? It requires these C libraries.

Then, could you share your `airone/settings.py`? It might contain some breaking changes on settings I guess.
For information, my `airone/settings.py` which I test it with my testing LDAP server is here:
```
AUTHENTICATION_BACKENDS = (
    'airone.auth.ldap.LDAPBackend',
)

AUTH_CONFIG = {
    'LDAP': {
        'SERVER_ADDRESS': 'ldap://localhost:389',  # not to use TLS for now
        'USER_FILTER': 'cn={username},ou=people,dc=nodomain'
    }
}
```
I also confirmed calling LDAP directly works fine:
```
$ ldapwhoami -x -D "cn=user,ou=people,dc=nodomain" -W
Enter LDAP Password:
dn:cn=user,ou=people,dc=nodomain
```

syucream:
@userlocalhost To dig the root cause more, could you tell me details of your environment with following confirmations?

### 1. Use no-TLS connection

Could you try to connect LDAP server without TLS, like using `ldap://<hostname>:389/`?

### 2. Skip certificate validation

Could you specify this option in `/etc/ldap/ldap.conf` to ignore certificate validation?
```
TLS_REQCERT   never
```

...Unfortunately I haven't been able to reproduce your situation yet, so I need more hints. As a similar case, I got such a error when I use LDAP over TLS with a wrong CN in my (self-signed) certificate.
```
[ERROR] asctime:2021-06-08 16:30:41,593 module:ldap     message:{'result': -1, 'desc': "Can't contact LDAP server", 'errno': 2, 'ctrls': [], 'info': 'TLS: hostname does not match CN in peer certificate'}       process:31668   thread:139638732171008
```
In this case, I passed the error if I specify `TLS_REQCERT   never` or modify CN. It might be relating to the error you got.

hinashi:
When I tried it in my development environment, I got the same error.
```
[ERROR] asctime:2021-06-14 10:18:06,737 module:ldap     message:{'result': -1, 'desc': "Can't contact LDAP server", 'ctrls': [], 'info': 'error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed (self signed certificate in certificate chain)'}      process:13503   thread:139783962650368
```

This is because the LDAP server is a self-signed certificate.
I solved it by adding the private key of the self-signed certificate to the certificate list specified in TLS_CACERT in /etc/ldap/ldap.conf.

userlocalhost (approved):
I understood this implementation works well in the proper environment. Thank you very much for digging into this problem and solving it, @hinashi. And thank you for your great work ,@syucream (sorry for making you trouble that is outside the scope of your responsibility).

userlocalhost:
I'll write this change to the changelog at https://github.com/dmm-com/airone/pull/152 after merging this PR.

PR #146, opened 2021-06-08
Fixed a search query timeout for long keywords

### LINKED ISSUE #145: Elasticsearch times out with simple search
All attributes are searched by simple search in the custom view.
Searching for a long keyword string times out the query to elasitcsearch.
```
hints = [{'name': x.name, 'keyword': query} for x in EntityAttr.objects.filter(is_active=True)]
Entry.search_entries(user, entity_ids, hints, or_match=True)
```
image

hinashi:
close: https://github.com/dmm-com/airone/issues/145

As a workaround, added an argument to search for all attributes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## In development
 ### Added
 * Added `^` and `$` operators on filtering `text` and `string` typed attributes in advanced search result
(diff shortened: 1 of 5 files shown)


userlocalhost (approved):
Thank you so much for fixing its problem!

PR #148, opened 2021-06-09
Version bytes decoding as UTF-8

takano32:


userlocalhost (commented):
Thank you for fixing. But this causes following problem.
https://app.circleci.com/pipelines/github/userlocalhost/airone/122/workflows/45e3b4d6-240a-4517-88ab-e3af32b08472/jobs/358

Instead of this commit, I fixed this problem in the #150.

PR #150, opened 2021-06-10
Fixed a minor problem about version displaying

userlocalhost:
This commit fixed a problem of showing AirOne version as a bytes typed value.


hinashi (approved):
LGTM!

PR #152, opened 2021-06-14
Fixed remaining work of #114 and #134

userlocalhost:
This commit has following changes which are remaining work of #114 and #134

* Remove unnecessary files which are added at #114
* Update changelog for adding feature in #114 and #134


hinashi (approved):
LGTM!

PR #154, opened 2021-06-21
Update INSTALLED_APPS to be able to handle human touched data

userlocalhost:
This commit added "[django.contrib.humanize](https://docs.djangoproject.com/en/2.2/ref/contrib/humanize/)" to the INSTALLED_APPS to be
able to handle data as a human touched one.


hinashi (approved):
LGTM!

PR #155, opened 2021-06-21
Expand Entry.to_dict to be able to more detail information

userlocalhost:
This commit expand Entry.to_dict method to be able to receive
with_metainfo parameter that would be passwd to the
AttributeValue.get_value method to be able to get more detail
information for AttributeValues of target Entry.


hinashi (approved):
LGTM!

PR #158, opened 2021-06-24
Fixed the result being different depending on the hint_attr_value of search_entries

hinashi:
The following PR added the hint_attr_value option to the search_entries function.
https://github.com/dmm-com/airone/pull/146

The search results were different because hint_attr_value is case sensitive.
Changed Elasticsearch query to regular expression format.

```
>>> Entry.search_entries(user, [entity.id], hint_attr_value='hoge')['ret_count']
3
>>> Entry.search_entries(user, [entity.id], hint_attr_value='Hoge')['ret_count']
3
```


userlocalhost (approved):
LGTM 👍

PR #159, opened 2021-06-29
Fixed a problem in the processing of import entry

userlocalhost:
This commit fixed a problem that occurs an Exception when an import job
is running at the situation that there is an abnormal Entry (e.g. there
are multiple Attributes of same name). And changed import processing to
finish job safely by making exception handler. Once an unexpected
problem might be happened in the import processing, this finishes its job
safely and records it that it's an error job.

## Execution result
Once I imported entry to an environment that there is an Entry that has multiple same named Attributes. I confirmed an Error was invoked at a Celery as below.
スクリーンショット 2021-06-29 12 03 52


hinashi (approved):
LGTM!
I confirmed it in the local environment.

PR #162, opened 2021-07-03
Install prettier as a JS code formatter

syucream:
[prettier](https://prettier.io/) is a JS code formatter that looks working for React/JSX well. So I would like to introduce it to improve our productivity in this PR. The most of changes in `frontend/` is modified by `$ npx prettier --write frontend/`


userlocalhost (approved):
Thank you very much!!

takano32 (approved):

PR #164, opened 2021-07-04
Refactoring React components

syucream:
Refactoring React components with focusing these points:
- Separate CREATE/UPDATE/DELETE buttons commonly used
- Bind prop values by `{var1, var2, ...}` to avoid redundant references
- Simplify map functions on JSX to unnest redundant lambda's


userlocalhost (approved):
LGTM. Thank you so much!!

PR #167, opened 2021-07-10
Split components and introduce fundamental test cases

syucream:
- Split some components fetches backend APIs
- Introduce some fundamental test cases with using [Jest](https://jestjs.io/ja/) and [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/)

Now we can run test cases with `$ npm run test`  and also its automatically executed by the CI yay! 🎉 

It contains https://github.com/dmm-com/airone/pull/162 and https://github.com/dmm-com/airone/pull/164 so please check these PRs before this.


userlocalhost (approved):
Thank you for your hard work!

PR #169, opened 2021-07-12
Changed not to create tokens with GET user.access_token API

hinashi:
Eliminating database updates with GET requests due to multi-database support.

Currently, when you open the user edit page with GET, an access token is created.
Removed the process of creating an access token.
If there is no access token, nothing will be displayed.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changed
 * Update Django to version 3.2.4 LTS (#153)
+* Changed not to create tokens with GET user.access_token API (#166)
+* Changed the behavior of token refresh
 
 ### Fixed
 * Fixed the result being different depending on the hint_attr_value of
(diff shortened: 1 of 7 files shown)


userlocalhost (commented):
Thank you very much for your work. It's great to decouple processing that send update requests to Databse from GET method request handler.

I have just one favor to ask of you to fix expressions of UI.

The "None" should be converted to other appropriate expression in a page that user see it because the that isn't a common noun but a technical one of Python.

For example,
  - In the "AccessToken" column, it would be better that "--- The token have not been issued yet. Please push 'Refresh' button to do it ---"
  - In the "date of expiry" column, "----/--/--" is better than "None"

スクリーンショット 2021-08-17 11 12 18

hinashi:
@userlocalhost 
Thank you for checking!

I have improved the UI.

- case: none token by user
image

- case: has exipre by user
image

- case: none exipre by user
image

- case: none token by admin
image

- case: has exipre by admin
image

- case: none exipre by admin
image

userlocalhost (approved):
LGTM👍 Thank you very much for fixing!

PR #172, opened 2021-07-17
Build React app in dev mode

syucream:
`production` build mode generates a minified .js without debuggable information. It's still under development, so it's better to use dev-mode.

```
# it generates .js that's suitable for debug
$ npm run build
# or
$ npm run build:development

# it generates .js that's suitable for production environment (e.g. its minified)
$ npm run build:production
```

### DIFF
--- a/templates/frontend/index.html
+++ b/templates/frontend/index.html
@@ -8,8 +8,6 @@
 <body>
   <div id="app" class="columns"><!-- React --></div>
 
-  <script>
-    {% include 'frontend/main.js' %}
-  </script>
+  <script src="/static/js/new-ui.js"></script>
 </body>
 </html>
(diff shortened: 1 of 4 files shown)


templates/frontend/index.html
```diff
@@ -8,8 +8,6 @@
 <body>
   <div id="app" class="columns"><!-- React --></div>
 
-  <script>
-    {% include 'frontend/main.js' %}
```
syucream, line 22: Built .js is something wrong in expanded html here. For now it refers a separated .js file.

takano32 (approved):
LGTM

userlocalhost (approved):
Good catch! Thank you!

PR #173, opened 2021-07-18
Reimplement job list features in React

syucream:
It reimplement job-list features in React, like this.
image

It contains #162, #164 and #167 so please check these PRs before this.

### DIFF
--- a/frontend/src/components/job/JobList.js
+++ b/frontend/src/components/job/JobList.js
@@ -1,0 +1,109 @@
+import {
+  List,
+  ListItem,
+  Table,
+  TableBody,
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableRow,
+} from "@material-ui/core";
+import Paper from "@material-ui/core/Paper";
+import Typography from "@material-ui/core/Typography";
+import Button from "@material-ui/core/Button";
+import { Link } from "react-router-dom";
+import React from "react";
+import { makeStyles } from "@material-ui/core/styles";
+
+const useStyles = makeStyles((theme) => ({
+  button: {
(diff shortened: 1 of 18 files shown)


frontend/src/components/job/JobList.js
```diff
@@ -0,0 +1,109 @@
...
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          {jobs.map((job) => (
+            <TableRow>
+              <TableCell>
+                <Typography>{job.target && job.target.name}</Typography>
+              </TableCell>
+              <TableCell>
+                <Typography>{job.operation}</Typography>
```
syucream, line 63: Let's aim it on another PR.

frontend/src/components/job/JobList.js
```diff
@@ -0,0 +1,109 @@
...
+              </TableCell>
+              <TableCell>
+                <Typography>{job.status}</Typography>
```
syucream, line 66: Let's aim it on another PR.

frontend/src/utils/AironeAPIClient.js
```diff
@@ -200,21 +200,8 @@ export function deleteGroup(groupId) {
   });
 }
 
-// FIXME implement internal API then call it
-export function getJobs() {
```
takano32, line 204: `limit` parameter is used as query parameter key more than often `nolimit`.

Use `limit` parameter  and if its value is `0` then HTTP requests often  processes the parameter having no limit as is also often has used.

How's it applying this way in this code?

Thanks for reading my comment.

userlocalhost (approved):
LGTM 👍 Thank you for your great work.

PR #174, opened 2021-07-19
fix urls

takano32:
Fix `urls` to avoid some warnings.

```
System check identified some issues:

WARNINGS:
?: (urls.W002) Your URL pattern '^/del/(\d+)$' [name='del_webhook'] has a route beginning with a '/'. Remove this slash as it is unnecessary. If this pattern is targeted in an include(), ensure the include() pattern has a trailing '/'.
?: (urls.W002) Your URL pattern '^/set/(\d+)$' [name='set_webhook'] has a route beginning with a '/'. Remove this slash as it is unnecessary. If this pattern is targeted in an include(), ensure the include() pattern has a trailing '/'.
?: (urls.W005) URL namespace 'entity' isn't unique. You may not be able to reverse all URLs in this namespace
```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,3 @@
 {
-    "python.pythonPath": ".venv/bin/python"
+    "python.defaultInterpreterPath": ".venv/bin/python"
 }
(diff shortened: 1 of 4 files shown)


userlocalhost (approved):
LGTM

userlocalhost (commented):
Would you please update Changelog for this commit?

PR #177, opened 2021-07-24
Reimplement import/export features in React

syucream:
Reimplement therese import/export features in React:
- import/export entities
- import/export entries
- import/export groups

https://user-images.githubusercontent.com/191684/126859815-c32087a8-a030-422e-80aa-ee6b679339a2.mov

It includes https://github.com/dmm-com/airone/pull/167 so please see it before this PR

### DIFF
--- a/frontend/src/components/common/ImportForm.js
+++ b/frontend/src/components/common/ImportForm.js
@@ -1,9 +1,9 @@
 import React, { useRef, useState } from "react";
-import AironeBreadcrumbs from "../components/common/AironeBreadcrumbs";
 import Typography from "@material-ui/core/Typography";
-import { Link } from "react-router-dom";
 import Button from "@material-ui/core/Button";
 import { makeStyles } from "@material-ui/core/styles";
+import PropTypes from "prop-types";
+import { useHistory } from "react-router-dom";
 
 const useStyles = makeStyles((theme) => ({
   button: {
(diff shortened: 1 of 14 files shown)


frontend/src/components/common/ImportForm.js
```diff
@@ -1,45 +1,41 @@
...
   },
 }));
 
-export default function Import({}) {
+export default function ImportForm({ importFunc, redirectPath }) {
   const classes = useStyles();
-  const [files, setFiles] = useState([]);
-  const filesRef = useRef(null);
+  const history = useHistory();
+  const [file, setFile] = useState([]);
```
takano32, line 17: Using `files` instead of `file` is looks good to me.

But above comment may be incorrect cause from my not good at React.

Thanks.

syucream: The current UI looks to accept only 1 file so I just follow that. If users will hope to import multiple files, we can support that on the new UI of course 👍 

@userlocalhost @hinashi any opinion?

userlocalhost: I think `file` is better for the perspective the current backend implementation. When AirOne supports multiple files, it would be better to change it to `files`.

userlocalhost (approved):
Thank you for your work. I confirmed all features, which are described in the description, were work well. 👍

PR #189, opened 2021-08-01
Roughly reimplement some pages for managing entries

syucream:
Roughly reimplement some remaining pages for managing entries

image

image

image

image

image

image

image

NOTE: Some pages just use dummy data without calling APIs

It includes #167 so please see it before this PR


frontend/src/components/entry/CopyForm.js
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from "react";
+import { copyEntry } from "../../utils/AironeAPIClient";
+import Button from "@material-ui/core/Button";
+import { useHistory } from "react-router-dom";
+import PropTypes from "prop-types";
+
+export default function CopyForm({ entityId, entryId }) {
+  const history = useHistory();
+
+  const [entries, setEntries] = useState("");
```
takano32, line 11: I guess `entries` is something like `[]` from variable name.
Is this, `""` is OK?

syucream: It should be ideally! Now I reuse current views receives string-typed `entries` so I make it a string value.
ref. https://github.com/dmm-com/airone/blob/aa743807236387aa1ce99ec56712caec23e0f77c/entry/views.py#L584

I want to resolve the issue on another PR 🙇

userlocalhost (commented):
I beg you to fix followings

### 1. Unable to edit attributes when creating Entry
There is no input element to edit attribute value of Entry at creating Entry page.
スクリーンショット 2021-09-30 16 14 00
スクリーンショット 2021-09-30 16 14 07

### 2. Showing Entity name instead of Entity-ID
Like above screenshot, breadcrumbs shows Entity-ID. I hope Entity name would be shown.

userlocalhost:
I made an issue for above comment and I'll merge this PR.

PR #190, opened 2021-08-03
Added selectors and fixed invalid HTML format

userlocalhost:
This commit adds some CSS selectors to be able to identify some HTML
elements, which describe job information. And fixed DOM tree to be
validated.


hinashi (approved):
LGTM!
Great to have tests of CSS selector.

PR #197, opened 2021-08-15
Implement list/get/create/update/delete operations for group

syucream:
- Implement GroupAPIV2 with DRF(Django Rest Framework)
- Implement frontend features with calling the API

We can generate OpenAPI based spec files with `drf_spectacular`,
```
$ python manage.py spectacular > spec.yaml
```
so we can show the specs on Swagger UI
image
and also we will get auto-generated client code with [openapi-generator](https://github.com/OpenAPITools/openapi-generator) from the spec.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,9 @@
 {
     "presets": [
-        "@babel/preset-env", "@babel/preset-react"
+        "@babel/preset-react",
+        [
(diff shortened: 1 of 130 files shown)


takano32 (approved):
LGTM

Especially using `drf_spectacular` is good, I think.

userlocalhost (approved):
LGTM

PR #200, opened 2021-08-17
Added job function that cannot be canceled

hinashi:
close: https://github.com/dmm-com/airone/pull/169

Defined a job that can be canceled with CAN_CANCEL_OPERATIONS.
Jobs that cannot be canceled have their buttons disabled.

image

### DIFF
--- a/job/models.py
+++ b/job/models.py
@@ -90,6 +90,14 @@ class Job(models.Model):
         JobOperation.NOTIFY_DELETE_ENTRY.value,
     ]
 
+    CAN_CANCEL_OPERATIONS = [
+        JobOperation.CREATE_ENTRY.value,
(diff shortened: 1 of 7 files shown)


job/models.py
```diff
@@ -90,6 +90,14 @@ class Job(models.Model):
         JobOperation.NOTIFY_DELETE_ENTRY.value,
     ]
 
+    CAN_CANCEL_OPERATIONS = [
```
userlocalhost: It would be better to change "CANCELABLE_OPERATIONS", I think.

userlocalhost (commented):
Thank you very much for your hard work.

job/models.py
hinashi: Thank you for your advice!
That name is better, so I'll change it.

PR #204, opened 2021-08-21
Reset default margin on the body

syucream:
Eliminate the default (from user agent stylesheet) margin.

# before

The `<body>` has `margin: 8px` automatically.

image

# after

image

### DIFF
--- a/templates/frontend/index.html
+++ b/templates/frontend/index.html
@@ -5,7 +5,7 @@
   <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&display=swap" />
   <title>Django DRF - React : Quickstart - Valentino G. - www.valentinog.com</title>
 </head>
-<body>
+<body style="margin: 0px">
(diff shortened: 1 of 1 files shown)


templates/frontend/index.html
```diff
@@ -5,7 +5,7 @@
   <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&display=swap" />
   <title>Django DRF - React : Quickstart - Valentino G. - www.valentinog.com</title>
 </head>
-<body>
```
takano32, line 8: If there is a CSS file that this file is reading, it is better to write it there if you can write it.

takano32: Sorry, there's not properly CSS file in this case.
Inline style sheet attribute is not bad.

takano32 (approved):
LGTM

userlocalhost (approved):
It's great!! Thank you!

PR #207, opened 2021-08-23
(DRAFT) Added feature to return default Attribute value

userlocalhost:
This commit adds a new feature to return default Attribute value without
making instance to DB for multiple Database measure.


hinashi (approved):
LGTM!
Thank you for your idea.
I will continue to work on it.

PR #218, opened 2021-08-27
Wrap django context served via window global

syucream:
Wrap `window.django_context` with `DjancoContext` class not to allow any components to access it directly. And also I add `User Cconfig` link on the header part with using the wrapped class.

image


userlocalhost (approved):
LGTM 👍

PR #219, opened 2021-08-28
Filter inactive entity attrs on entry-export

syucream:
entry-export can involve inactive entity attributes. For e.g. a user do this steps,:
1. create an entity with an attribute
2. create an entry
3. delete the entity-attribute
4. export the entry

The exported file contains a deleted entity attribute. For e.g.:
```yaml
entry100:
- attrs:
    a1_deleted_20210828_123830: entry100-1
  name: entry100-1
- attrs: {}
  name: entry100-2
```

This is because entry-export doesn't check if entity-attribute is active, so I would like to introduce checking the active flag to make it natural.

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -1238,7 +1238,7 @@ class Entry(ACLBase):
         # that are added after creating this entry.
         self.complement_attrs(user)
 
-        for attr in self.attrs.filter(is_active=True):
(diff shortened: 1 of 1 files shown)


hinashi (approved):
LGTM!

takano32 (approved):
LGTM

PR #220, opened 2021-08-28
Catch ValueError caused by yaml.load()

syucream:
If importing an invalid date value, current import feature failed by uncaught `ValueError`. For e.g.

image
image

So I fix the importer to catch `ValueError` and return 400 error.


hinashi (approved):
LGTM!

takano32 (approved):
LGTM

PR #222, opened 2021-08-28
Deny accesses to a deleted user

syucream:
Now the user management page receives requests associated to a deleted user. Its better to deny such accesses.


userlocalhost (approved):
Thank you very much for your work. I confirmed I can't changed user password, which has already been deleted.

PR #170, opened 2021-07-12
Changed the HTTP method on the entry.export page from GET to POST

hinashi:
Eliminating database updates with GET requests due to multi-database support.

Currently, the entry export task is running in the background job.
Creating a job updates the database.
I changed from the GET method to the POST method.

image

PR #225, opened 2021-09-01
Changed some function to use multiple databases

hinashi:
Changed below to use multiple databases.

- Added the django-replicated library
- Removed my own DB selection logic
- Changed the logout from GET to POST method
- Removed complement_attrs when request show entry page
- Added API to create Attribute for edit ACL
- Added feature to return default Attribute value


entry/models.py
```diff
@@ -1116,26 +1092,42 @@ def complement_attrs(self, user):
...
             if attr.values.exists():
-                last_value = attr.get_latest_value()
+                last_value = attr.get_latest_value(is_readonly=True)
+                if last_value is None:
+                    ret_attrs.append(attrinfo)
+                    continue
+
                 if not last_value.data_type:
                     last_value.data_type = attr.schema.type
                     last_value.save()
```
userlocalhost: DB への書き込みが発生しそうな予感...
仕様変更（get_available_attrs で DB への更新を行わない）に伴い、DB へのアクセスを伴う処理は無くした方が良いかと思います。
（当該ロジックは通らない想定なので、単純に消すだけでも問題ないかと思います）

airone/auth/view.py
```diff
@@ -0,0 +1,14 @@
+from django.contrib.auth import logout as django_logout
+from django.views.decorators.csrf import csrf_protect
+from django.http import HttpResponse
+
+from airone.lib.http import render
+
+
+@csrf_protect
```
takano32, line 8: 👍

userlocalhost:
Would you please correct this processing? Because it might be run from GET request event handler (history method, which is declared at entry/views.py)
https://github.com/dmm-com/airone/blob/9c535afc2b946ea7f95b33ac841ad9ba520498fa/entry/models.py#L178-L181

userlocalhost:
Those following methods in `Entry.complement_attrs()` might send queries to DB for update. And this method is called by `export()` method, which expects not to send update queries to DB. Could you please caring them?

https://github.com/dmm-com/airone/blob/9c535afc2b946ea7f95b33ac841ad9ba520498fa/entry/models.py#L1074

https://github.com/dmm-com/airone/blob/9c535afc2b946ea7f95b33ac841ad9ba520498fa/entry/models.py#L1087

https://github.com/dmm-com/airone/blob/9c535afc2b946ea7f95b33ac841ad9ba520498fa/entry/models.py#L1093

hinashi:
@userlocalhost 
Thank you for checking.
I forgot that export was changed from GET to POST in the PR below.
https://github.com/dmm-com/airone/pull/170

Changed entry model export from database read to write.

hinashi:
@userlocalhost 
I removed the DB write logic from get_value_history.
The data_type is always set, so it looks like unnecessary logic.

In addition, I deleted the unused DB selection logic because it remained.

userlocalhost:
Thank you very much for your hard work!

PR #226, opened 2021-09-04
Try to show/record entry history

syucream:
Current entry-history page doesn't provide updates on the entry itself -- in other words, it actually supports only updated on their attributes. So I would like to support to record activities of entry(create/edit/delete/restore).

We can realize it with current data structure, like below:
image

### DIFF
--- a/entry/tasks.py
+++ b/entry/tasks.py
@@ -337,6 +337,11 @@ def restore_entry(self, job_id):
         # update entry information to Elasticsearch
         entry.register_es()
 
+        # register operation History for restoring entry
+        user = User.objects.filter(id=job.user.id).first()
+        entry_history = user.seth_entry_mod(entry)
+        entry_history.mod_entry(entry, 'restore deleted entry')
+
         # update job status and save it
         job.update(Job.STATUS['DONE'])
 
--- a/entry/views.py
+++ b/entry/views.py
@@ -25,7 +25,7 @@ from airone.lib.profile import airone_profile
 from entity.models import Entity
 from entry.models import Entry, Attribute, AttributeValue
 from job.models import Job, JobOperation
-from user.models import User
+from user.models import History, User
 from group.models import Group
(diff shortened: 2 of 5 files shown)


entry/views.py
```diff
@@ -290,12 +293,18 @@ def do_edit(request, entry_id, recv_data):
...
     entry.name = recv_data['entry_name']
 
     # set flags that indicates target entry is under processing
     entry.set_status(Entry.STATUS_EDITING)
     entry.save()
 
+    # register operation History for editing entry
+    if prev_name != recv_data['entry_name']:
+        h = user.seth_entry_mod(entry)
+        h.mod_entry(entry, 'old name: "%s"' % prev_name)
```
syucream, line 306: Should we always record updates with including updated on attributes here? I think it'll be noisy so I filter it once.

entry/tasks.py
```diff
@@ -337,6 +337,11 @@ def restore_entry(self, job_id):
         # update entry information to Elasticsearch
         entry.register_es()
 
+        # register operation History for restoring entry
+        user = User.objects.filter(id=job.user.id).first()
+        entry_history = user.seth_entry_mod(entry)
+        entry_history.mod_entry(entry, 'restore deleted entry')
```
syucream, line 343: Is it better to have another operation value like `RESTORE_ENTRY` then use it here?

hinashi (approved):
LGTM!
You may want ​to add a test.
https://github.com/dmm-com/airone/blob/6244e1f48fb524b87aaaab322d26d47b51a0fcdf/user/tests/test_model.py#L45

userlocalhost (approved):
Thank you very much for your hard work!!

Would you like to replacing entry history and attribute history.
スクリーンショット 2021-09-30 16 28 24

syucream:
Use https://django-simple-history.readthedocs.io/en/latest/ instead.

PR #231, opened 2021-09-08
Fixed some attributes are not updated in advanced search results

### LINKED ISSUE #230: Some attributes are not updated in advanced search results
Some attributes are not updated in advanced search results.
No data was registered in ElasticSearch.

(e.g.
image
image

hinashi:
close: https://github.com/dmm-com/airone/issues/230

Fixed some attributes are not updated in advanced search results.
Updating the entry after the fix will resolve it.

image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Fixed
 * Fixed the problem that the URL of Webhook API is different (#202)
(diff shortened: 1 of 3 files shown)


userlocalhost (approved):
Thank you very much for hard work to dig into the reason of this problem 👍

PR #234, opened 2021-09-11
Use import-js/eslint-plugin-import

syucream:
Install https://github.com/import-js/eslint-plugin-import to perform auto-ordering import statements. Now we can `$ npm run fix` to apply the linter rule.

NOTE its sometimes wrong, like:
```
...
  22:1  error  `./pages/OperationHistory` import should occur before import of `./pages/SearchResults`  import/order

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
...
```
but its recoverable easily with re-running `$ npm run fix`.

ref. https://buildersbox.corp-sansan.com/entry/2021/05/28/110000

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,37 @@
+module.exports = {
+  plugins: ['import'],
+  rules: {},
+  parserOptions: {
+    sourceType: "module",
+    ecmaVersion: 11,
+    ecmaFeatures: {
+      jsx: true,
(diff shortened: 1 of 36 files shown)


userlocalhost:
I'm so sorry that I couldn't resolve this confliction because it's too complex. So I created another PR (#266) that includes this change.

PR #235, opened 2021-09-11
Stop using export-default

syucream:
`export default` looks like a not smart way.

refs.
- https://typescript-jp.gitbook.io/deep-dive/main-1/defaultisbad
- https://engineering.linecorp.com/ja/blog/you-dont-need-default-export/


userlocalhost:
https://github.com/dmm-com/airone/pull/265 overwrite this commit

PR #237, opened 2021-09-12
Implement creating/updating a user

syucream:
Implement creating/updating a user and updating a password which's missed.


hinashi (approved):
LGTM!

PR #238, opened 2021-09-14
Fixed the session being cut off after logging in with multi-DB

hinashi:
The django_replicated library selects SlaveDB for GET requests, and MaserDB otherwise.
Also, for the redirect after POST, select MasterDB even for GET.
This is only once. Subsequent GET requests will select SlaveDB.

If SlaveDB has replication delays, referencing non-existent data can be a problem.
e.g.) there is no session data in SlaveDB immediately after logging in.

After selecting MasterDB, I changed it so that MasterDB is selected until cokkie expires.

The original code is below.
https://github.com/yandex/django_replicated/blob/125ec732162091411811cf77c8061e4ed6fcd7de/django_replicated/middleware.py#L142-L155

### DIFF
--- a/airone/lib/db.py
+++ b/airone/lib/db.py
@@ -1,0 +1,27 @@
+import logging
+
+from django_replicated.middleware import ReplicationMiddleware
+from django_replicated.utils import routers
+
+log = logging.getLogger('django_replicated.middleware')
+
+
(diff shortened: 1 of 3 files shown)


userlocalhost (approved):
It's brilliant! Thank you very much for finding out great workaround solution. 👍

PR #245, opened 2021-09-17
Fixed displaying old values on the show entry page

hinashi:
In rare cases, multiple Attribute may be created.
Fixed displaying old values on the show entry page.


userlocalhost (approved):
LGTM 👍 Thank you for fixing a bug!

PR #248, opened 2021-09-18
Escape \ on ES query

syucream:
If filter condition has `\`, ES sometimes returns 400 error with saying `elasticsearch.exceptions.RequestError: TransportError(400, 'search_phase_execution_exception', 'failed to create query:...`. So I think its better to escape `\` before sending the query.

before:
image

after:
image

### DIFF
--- a/airone/lib/elasticsearch.py
+++ b/airone/lib/elasticsearch.py
@@ -261,12 +261,15 @@ def _get_hint_keyword_val(keyword: str) -> str:
     Returns:
         str: If a character corresponding to the empty string specified by CONFIG is entered,
             the empty character is returned.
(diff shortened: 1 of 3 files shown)


syucream:
btw, the single `\` is handled as empty query with https://github.com/dmm-com/airone/blob/3c807b54b35c8c651b03c84c1a6bc4ad756e8911/entry/settings.py#L11 . Why does it that? Is it reasonable to escape `\` on this PR?

syucream:
Just for a workaround, I replace keyword containing backslash with ' ' to sanitize it.

hinashi (approved):
LGTM!

userlocalhost (approved):
LGTM 👍 It's "Done is better than perfect".

PR #249, opened 2021-09-21
support for multiple databases

### LINKED ISSUE #166: About the load bottleneck of the Master database
## Background
AirOne is currently sending all DB queries to the master database.
As the number of users increases, the performance of the master database becomes a bottleneck.
The select queries should be sent to the slave database for reduce the load on the master database.

However, there is a problem sending all select queries to the slave database.
AirOne's database is MySQL.
The master / slave configuration is thinking of MySQL replication.
MySQL replication can cause replication delays and should be considered.
When databases is delay, A problem occurs a process called SELECT (delay) → UPDATE.

## Solution
Only the GET method uses the slave database.
The reason is that the GET method has no update process.
However, AirOne has a process to update with the GET method and needs to be fixed.

| | DB |
| ---- | ---- |
| GET method | Slave |
| POST method | Master |
| Celery job | Master |

hinashi:
close: https://github.com/dmm-com/airone/issues/166

https://github.com/dmm-com/airone/pull/207
https://github.com/dmm-com/airone/pull/225
https://github.com/dmm-com/airone/pull/238
https://github.com/dmm-com/airone/pull/245


userlocalhost (approved):
Thank you for your hard work. It must be great progress of our project 👍

PR #258, opened 2021-09-24
Added param of editting user for ldap

### LINKED ISSUE #256: No UI to switch to LDAP authentication
It has an LDAP authentication function, but it cannot be used because
there is no UI to switch from Django password authentication to LDAP authentication.

(c.f. https://github.com/dmm-com/airone/blob/master/airone/auth/ldap.py）

hinashi:
close: https://github.com/dmm-com/airone/issues/256

Added a UI to switch from Django password authentication to LDAP authentication.
Changed the time notation of tokens.
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## In development
 
+### Added
+* Added param of editting user for ldap (#256)
+
 ### Changed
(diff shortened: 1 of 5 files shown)


userlocalhost (approved):
LGTM 👍 
I confirmed that I could change the way of authentication from local DB to LDAP.

PR #259, opened 2021-09-24
Changed cookie of session to secure attribute, and to return HSTS header

### LINKED ISSUE #257: About secure HTTPS settings
AirOne is an information management system, so it needs to be a secure application.
The following are measures against session hijacking.
- Enable the secure attribute of session cookies.
- Set the HSTS header.

hinashi:
close: https://github.com/dmm-com/airone/issues/257

image

image


userlocalhost (approved):
LGTM 👍

PR #260, opened 2021-09-25
Redirect authenticated users to the top page

syucream:
When authenticated users access to the login page, it simply shows the login form without redirect. Its better to redirect them to the default page.


hinashi (approved):
LGTM!
I think it's a great implementation.

userlocalhost (approved):
LGTM 👍 Good Job!!

PR #263, opened 2021-09-27
Fixed that the entry being created cannot be deleted

### LINKED ISSUE #242: Same name entry might be created by restoring processing
There is no preventive measures not to create same name entries by restoring processing.

## Reproducing progress

1. Create an Entry (name: test1) at arbitrary Entity (e.g. Entity1).
2. Delete Entry of test1, which is created at (1).
3. Create another same name Entry (test1) at same Entity (Entity1).
4. Delete Entry of test1, which is created at (3).
5. Restore both Entries, which are deleted at (2) and (4) simultaneously.

After doing (5), multiple Entries, which are named test1, are restored in rare case.

## Cause of this bug
Once later request (which is sent by procedure(3)) was passed following checking before the former request (which is sent by procedure(2)) was done by celery,
https://github.com/dmm-com/airone/blob/master/entry/views.py#L680-L685

Both requests make restore jobs as below. And both requests are independent. Therefore, same name Entries might be created by this processing.
https://github.com/dmm-com/airone/blob/master/entry/views.py#L690

hinashi:
close: https://github.com/dmm-com/airone/issues/242

There is a problem that the entry is not deleted if it is deleted while creating the entry.
This can result in duplicate entries with the same name.
It can be reproduced by the following method.

1. Create entry
2. Delete entry
3. Create entry with the same name

I think https://github.com/dmm-com/airone/issues/242 is the same problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Fixed
 * Fixed the problem that the URL of Webhook API is different (#202)
 * Fixed some attributes are not updated in advanced search results (#230)
+* Fixed that the entry being created cannot be deleted (#242)
 
 ## v3.2.0
 
--- a/acl/models.py
+++ b/acl/models.py
@@ -45,11 +45,11 @@ class ACLBase(models.Model):
 
     def set_status(self, val):
         self.status |= val
-        self.save()
+        self.save(update_fields=['status'])
 
     def del_status(self, val):
         self.status &= ~val
-        self.save()
+        self.save(update_fields=['status'])
 
     def get_status(self, val):
         return self.status & val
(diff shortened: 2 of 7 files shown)


userlocalhost:
Thank you very much for digging into this problem and finding out this reproducibility👍 

I confirmed this problem would be happened by your reproduction procedure. But I also confirmed same problem was happened after applying this branch.
スクリーンショット 2021-10-04 11 33 18
スクリーンショット 2021-10-04 11 44 42

hinashi:
@userlocalhost 
Thank you for your check and fix!

userlocalhost (approved):
Thank you very much for fixing very difficult bug! I really appreciate it!!

PR #264, opened 2021-09-28
Refactored the entry list page

hinashi:
Separated the search function to add its own search function to the custom view.
Does not affect operation.


userlocalhost (approved):
Thank you for your work 👍

PR #266, opened 2021-09-28
Applied eslint reorganization for every code

userlocalhost:
This inherits https://github.com/dmm-com/airone/pull/234


hinashi (approved):
LGTM!

Issue #272, opened 2021-10-07
(BUG) AttributeValue.parent_attr might be different with actual one


userlocalhost:
When an Entry that has array_xxx typed Attribute was cloned, co-AttributeValue's parent_attr member is different with AttributeValue.parent_attr.

PR #273, opened 2021-10-07
Fixed bug of cloning Entry that has array_xxx typed Attribute

userlocalhost:
Close: https://github.com/dmm-com/airone/issues/272


hinashi (approved):
LGTM!
I confirmed the fix in my local environment.

PR #274, opened 2021-10-09
Centralize routing paths

syucream:
Currently it has many hardcoded paths anywhere that makes hard to modify routing config. I would like to centralize these paths to Routes.js to improve maintenanceability.


userlocalhost (approved):
LGTM 👍

PR #275, opened 2021-10-09
Add some test cases with React Testing Library experimentally

syucream:
SSIA


userlocalhost (approved):
It's "Done is better than perfect".

PR #277, opened 2021-10-12
Implmented entry.api_v2 for requests from react

userlocalhost:


userlocalhost:
## ToDo
Enable to show Entry information getting through the getEntry().

PR #279, opened 2021-10-14
Fixed unused URL settings

### LINKED ISSUE #278: There is an unused URL setting
The following URL is not used, so it is better to remove it.
- http://localhost:8080/admin/
- http://localhost:8080/api-auth/
- http://localhost:8080/entity/auth/

hinashi:
close: https://github.com/dmm-com/airone/issues/278


userlocalhost (approved):
LGTM 👍 Good catch!!

PR #281, opened 2021-10-16
Add missing null-checks on deleted-list-entry page

syucream:
The deleted-list-entry page misses some null-checks that occurs JavaScript errors. So I fix it to show available attributes on the page anyway.

image


hinashi (approved):
LGTM!

userlocalhost (approved):
LGTM 👍 Thank you for your work!!

PR #285, opened 2021-10-23
Catch ScannerError on importing entry

syucream:
Some yaml files may occur uncaught `yaml.scanner.ScannerError` on importing entry.

image

It's better to catch the error and show reason of the error.

### DIFF
--- a/entry/tests/fixtures/invalid_import_data4.yaml
+++ b/entry/tests/fixtures/invalid_import_data4.yaml
@@ -1,0 +1,5 @@
+# NBSP, instead of whitespace
+hoge:
+  - name: entry
+    attrs:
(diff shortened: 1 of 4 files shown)


entry/tests/fixtures/invalid_import_data4.yaml
```diff
@@ -0,0 +1,5 @@
+# NBSP, instead of whitespace
+hoge:
+  - name: entry
```
syucream, line 3: It contains *non-breaking space*, not whitespace occurs ScannerError.

hinashi (approved):
LGTM!

userlocalhost (approved):
LGTM 👍

PR #286, opened 2021-10-23
Attach referring entries on yaml export

syucream:
It allows us to get list of referrals on advanced search page.

If we search entries with `has_referral`:
image
the yaml exporter provides referral info on `.<entity>[index].referrals` element:
```yaml
referred:
- attrs:
    a1: a1
  name: referred1
  referrals:
  - entitiy: referring
    entry: referring1
```


hinashi (approved):
LGTM!
I confirmed that there is no effect on the import.

userlocalhost (approved):
LGTM 👍 It's an elegant code.

PR #288, opened 2021-10-25
Added flag variable for E2E test

userlocalhost:
This commit adds variable that is used for E2E test to be able to handle
event registration processing is over.


hinashi (approved):
LGTM!
Thank you for your support.

PR #291, opened 2021-10-28
Wrote test for entry/AttributeValue.js using Enzyme

userlocalhost:
This commit adds tests for some test cases using Enzyme, which is
another testing framework for React application.
(c.f. https://enzymejs.github.io/enzyme/)


userlocalhost:
https://github.com/dmm-com/airone/pull/356 inherits core concept of this PR. So it should be closed.

PR #294, opened 2021-10-30
Fix some remaining warnings

syucream:
SSIA


hinashi (approved):
LGTM
I confirmed that there is no warning on the changed page.

PR #295, opened 2021-11-01
Fixed no permission check in advanced search

hinashi:
Changed the following in advanced search.

image

- Changed to save permission information in ElasticSearch.
(Is_public, default_permission for Entry and Attribute)
- Enabled to select Entity and Attribute without permission.
- Sorted Entity and Attribute choices.
- Items for which not have permission are not displayed.
Also, the background color has been changed.

ElasticSearch data needs to be updated before review.
```
python tools/update_es_document.py
```


entry/models.py
```diff
@@ -1459,28 +1469,132 @@ def search_entries(kls, user, hint_entity_ids, hint_attrs=[], limit=CONFIG.MAX_L
...
+
+            # Check for has permission to EntityAttr
+
+            for index, hint_attr in enumerate(hint_attrs):
+                if 'name' not in hint_attr:
+                    continue
+
+                hint_entity_attr = entity.attrs.filter(name=hint_attr['name'],
+                                                       is_active=True).first()
+                hint_attrs[index]['permission'] = True if (
```
userlocalhost: You can refactor this processing as below.
* current: hint_attrs[index]['permission']
* fixed: hint_attr['permission]

entry/models.py
```diff
@@ -1459,28 +1469,132 @@ def search_entries(kls, user, hint_entity_ids, hint_attrs=[], limit=CONFIG.MAX_L
...
+            'ret_values': [],
+        }
+        for hint_entity_id in hint_entity_ids:
+            # Check for has permission to Entity
+            entity = Entity.objects.filter(id=hint_entity_id, is_active=True).first()
+            if not (entity and user.has_permission(entity, ACLType.Readable)):
+                continue
+
+            # Check for has permission to EntityAttr
+
```
userlocalhost: Please erase this blank line.

airone/lib/elasticsearch.py
```diff
@@ -105,8 +110,16 @@ def recreate_index(self) -> None:
                                 'referral_id': {
                                     'type': 'integer',
                                     'index': 'false',
+                                },
+                                'permission': {
+                                    'type': 'boolean',
+                                    'index': 'true',
```
userlocalhost: You may want to change this variable name to compatible with following name convention for boolean type variable.
(c.f. https://qiita.com/GinGinDako/items/6e8b696c4734b8e92d2b)

Candidate is
 - is_readable
 - is_permitted

userlocalhost:
In a simple search processing, It's not necessary to check permissions of EntityAttr and Attribute I think. By this change, processing time could be reduce.

userlocalhost:
I confirmed each advanced search and simple search works well 👍Thank you for your hard work!!
スクリーンショット 2021-11-15 17 21 57

If you have leeway to brush it up, would you please fix following small glitch?

* In the simple work, following exception will be occurred when uesr specifies '999999999999999999999999999'. You may want to add validator for boundary value.
  スクリーンショット 2021-11-15 17 42 51

hinashi:
Thank you for confirmed.
Added validation because an error occurred when a parameter larger than max_result_window was specified.

error
```
TransportError(500, 'search_phase_execution_exception', 'Result window is too large, from + size must be less than or equal to: [500000] but was [500100]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.')
```

PR #296, opened 2021-11-06
Apply async/await styles on pages/

syucream:
[async/await](https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Statements/async_function) is a modern syntax to describe asynchronous process without redundant promise chaining. I prefer it rather than using promise.

For now, I try to rewrite page components with async/await. How about it? If you'll feel it good, then I will apply the change to remaining component.


userlocalhost (commented):
It seems good to me 👍 Thank you for your great work!
But I'm afraid that I might not catch up with your true intension. So, would you please consult with you about it when I could meet you?

PR #305, opened 2021-11-06
Separate Table with pagination as a common component

syucream:
We have some `<Table>` with pagination so I would like separating the common logic as common component `<PaginatedTable>`


userlocalhost (approved):
LGTM 👍 Thanks for your tough work!!

PR #306, opened 2021-11-06
Manage ACL of entity/entry in new-ui

syucream:
It's now unimplemented so I would like to support it in new-ui too. It supports to edit ACLs of entity/entry.

image


userlocalhost (approved):
LGTM 👍

PR #309, opened 2021-11-15
Fixed a different number of entries displayed on the entity dashboard

### LINKED ISSUE #308: Different number of entries displayed on entity dashboard
Advanced search results and entity dashboard numbers are different.

The cause is that the filter condition is in the is_active status of the attribute.
This is resolved by adding the is_active status of the entry to the filter condition.

However, this is a rare case because the attribute must be removed when the entry is deleted.

https://github.com/dmm-com/airone/blob/2e453edf10c9f1dcde30c634fee4bb91bdc65b02/entity/views.py#L355-L360

https://github.com/dmm-com/airone/blob/2e453edf10c9f1dcde30c634fee4bb91bdc65b02/entry/models.py#L1207-L1222

hinashi:
close: https://github.com/dmm-com/airone/issues/308

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 * Fixed an issue that caused redirects by incorrect URL links
+* Fixed a different number of entries displayed on the entity dashboard (#308)
 
 ## v3.3.1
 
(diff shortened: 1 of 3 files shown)


userlocalhost (approved):
LGTM 👍

PR #310, opened 2021-11-15
Expand tools for elasticsearch

userlocalhost:
This commit expands tools for elasticsearch registration to be able to
specify which Entities will be registered onto the elasticsearch.

スクリーンショット 2021-11-15 16 24 55


hinashi (approved):
LGTM
It's very convenient!

PR #312, opened 2021-11-24
Fixed some issues with simple search and advanced search

hinashi:
- Fixed an issue where regular expressions would fail in advanced searches.
  - before
image
  - after
image

- Fixed couldn't AND and OR searches in simple search
  - before
image
(search 『ho&ge』)
  - after
image

- Added CSS selector for ui-test in advanced search


userlocalhost (approved):
I confirmed all degraded features, which are related with search processing, are fixed. Thank you very much!!

PR #313, opened 2021-11-26
Changed the SSO authentication method from oauth to saml

hinashi:
Changed the authentication method from OAUTH to SAML,
So that it can be supported by other than google too.


userlocalhost (approved):
LGTM 👍

syucream (approved):
LGTM

PR #318, opened 2021-12-04
Enable to preserve DB data for MySQL on docker compose

syucream:
I would like to use docker compose volume to preserve MySQL DB data in my local env.


userlocalhost (approved):

hinashi (approved):
LGTM!

PR #322, opened 2021-12-06
Fixed an issue where advanced search narrow down was slow

### LINKED ISSUE #321: Narrowing down advanced search results may take some time
Narrow down of advanced search result calls the entry search API with ajax.
Entry search API outputs all attributes.
But advanced search results require only the specified attributes.
There is a problem that the output of all attributes is slow.

image

hinashi:
close: https://github.com/dmm-com/airone/issues/321

Added is_output_all option to entry search API.
In the advanced search result narrowing down, is_output_all was set to False to solve the problem.
image


userlocalhost (approved):
Thank you for your good work!!

Issue #324, opened 2021-12-07
Different count of ret_values in advanced search results


hinashi:
Count of advanced search results displayed is 100.
However, if you search for multiple entities, 100 items may not be displayed.

PR #325, opened 2021-12-07
Fixed/advanced search/old param

### LINKED ISSUE #323: Error when specifying old parameters in advanced search
There was a problem with the following PR.
https://github.com/dmm-com/airone/pull/295

attr[] and has_referral are targeted.
image
image

hinashi:
close: #323 #324 
image

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -186,12 +186,15 @@ def advanced_search_result(request):
     if not recv_attr and not attrinfo:
         return HttpResponse("The attr[] or attrinfo parameters is required", status=400)
 
-    # build hint attrs from JSON encoded params,
-    # or attr[] the older param to keep backward compatibility
+    # The "attr" parameter guarantees backward compatibility.
+    # The "atterinfo" is another parameter,
+    # that has same purpose that indicates which attributes to search,
+    # And "attrinfo" is prioritize than "attr".
     # TODO deprecate attr[]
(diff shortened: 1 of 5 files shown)


dashboard/views.py
```diff
@@ -169,18 +169,32 @@ def advanced_search_result(request):
...
     attrinfo = request.GET.get('attrinfo')
     entry_name = request.GET.get('entry_name')
 
+    # check referral params
+    # process of converting older param for backward compatibility
+    if has_referral == 'true':
+        has_referral = ''
+    if has_referral == 'false':
+        has_referral = False
+
```
userlocalhost, line 184: This must be a workaround processing. We have to refactor this processing by implementing new parameter that narrow down referral entries in near future.

dashboard/views.py
```diff
@@ -169,18 +169,32 @@ def advanced_search_result(request):
     user = User.objects.get(id=request.user.id)
 
     recv_entity = request.GET.getlist('entity[]')
+    recv_attr = request.GET.getlist('attr[]')
```
userlocalhost, line 172: Please add following message in comment.
```
# The "attr" parameter guarantees backward compatibility. The "atterinfo" is another parameter that has same purpose that indicates which Attributes to search. And "attrinfo" is prioritize than "attr". 
```

hinashi: Thank you for checking, added a comment.

Issue #327, opened 2021-12-07
Exception error when specifying an invalid parameter in advanced search


hinashi:
## Error1
If you specify a parameter of a different type, an exception error will occur.

- in entry search API
  - `"entities": [{"key": "value"}]`
  - `"attrinfo": [{"key": "value"}]`
  - `"attrinfo": [{"name": []}]`
  - `"entry_name": ["hoge"]`

- in dashboard/advanced_search_result page
  - `entity[]=[]`
  - `attrinfo=[{"key": "value"}]`
  - `attrinfo=[{"name": []}]`

## Error2
An error occurs when you specify `\^` for has_referral in advanced search.

Issue #326, opened 2021-12-07
Fixed workaround processing on advanced search


userlocalhost:
We have to solve following workaround.
(c.f. https://github.com/dmm-com/airone/pull/325#discussion_r763585337)

PR #328, opened 2021-12-07
Fixed/advanced search/exception

### LINKED ISSUE #314: [API] expand '/api/v1/entry/search' to be able to get results that are exactly matched
Current its API handler returns results that are partially matched with user specified `attrinfo`. However, in some use-cases, user want to get results that are exactly matched ones.

It might be better to implement it by adding new flag (like `exact_match`) in the `attrinfo` parameter.

hinashi:
close: #314 #326 #327 #330

- Fixed the order of entities when is_all_entities is specified in advanced search
(#330)
image

- Fixed not being able to use regexp in entry names in the entry search AP
(#314)
image

- Refactored referral param in advanced search
(#326)

- Fixed an exception error when specifying an invalid parameter in advanced search
(#327)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+* Fixed not being able to use regexp in entry names in the entry search API (#314)
 
 ## v3.4.0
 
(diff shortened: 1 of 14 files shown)


userlocalhost (approved):
LGTM 👍

hinashi (commented):
Thank you for your check and fix!

PR #334, opened 2021-12-17
Updated Django version that fixed security bug (CVE-2021-44420)

userlocalhost:
This commit change dependent Django version from v3.2.5 to v3.2.10 that
applied a patch of CVE-2021-44420.
(c.f. https://www.djangoproject.com/weblog/2021/dec/07/security-releases/)


hinashi (approved):
LGTM!

PR #341, opened 2021-12-20
Fixed issues where referrals are exported without add_referral

hinashi:
Fixed a bug caused by the following PR
https://github.com/dmm-com/airone/pull/328


userlocalhost (approved):
LGTM 👍

PR #343, opened 2021-12-20
Fixed an issue with array type attributes when copying entries

### LINKED ISSUE #342: There is a problem with the array type attribute when copying an entry
Reproduce in the following flow.
1. Create Ref Entry
2. Create Entry with array object type attribute set to ref entry
3. Copy Entry
4. Change AttributeValue of array_object type in copy entry
5. Check reference entry of Ref Entry

hinashi:
close: #342


userlocalhost (approved):
Thank you very much!! It's a great bug fix 👍

PR #347, opened 2021-12-27
Added to be able to insert custom javascript

hinashi:
There are cases where you want to add javascript to every page.
For example, Google Analytics, Microsoft Clarity.
Added to be able to insert custom javascript.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -41,7 +41,7 @@
     </div>
 
     <!-- custom script -->
-    {% include 'custom_script.html' %}
+    {% include 'custom_view/custom_script.html' %}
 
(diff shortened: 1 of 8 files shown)


templates/base.html
```diff
@@ -40,6 +40,9 @@
       </div>
     </div>
 
+    <!-- custom script -->
+    {% include 'custom_script.html' %}
```
userlocalhost: It would be better to set 'custom_view/custom_script.html'.

hinashi: Thank you for your advice!

userlocalhost:
LGTM 👍 Thank you for fixing.

PR #353, opened 2022-01-04
Fixed take a long time to create entry

hinashi:
Only when creating an ACLBase object Changed to create permissions.

Receive object change notifications with django signals.
https://docs.djangoproject.com/en/3.2/topics/signals/

For my development environment (about 2.5 million ACLBase objects)
Entry creation has been reduced from 5 seconds to less than 1 second.

Not only when creating entries, but also the time for changes has been reduced.


userlocalhost (approved):
It's awesome!! Thank you for making this great feature!!!

PR #357, opened 2022-01-08
Try using auto-generated client by openapi-generator

syucream:
Use auto-generated, well-typed REST API client code with [openapi-generator](https://github.com/OpenAPITools/openapi-generator) and OpenAPI spec file generated by [drf-spectacular](https://github.com/tfranzel/drf-spectacular).

First of all, we started using TypeScript to ensure our React components type-safe. But now API client code is not so type-safe. For e.g. on getting entity, we do:
```typescript
const resp = await fetch("/path/to/api/v2/getting/entity");
return await resp.json();
```
Unfortunately these API returns untyped values, so type checking should not work well.

We can get OpenAPI spec file with drf-spectacular. The spec contains data types originally defined in DRF serializer class. Then we also can generate data types in TypeScript with openapi-generator based on the spec file. As the result, we can ensure TS implementation type-safe with Python code.

NOTE: openapi-generator will not be a so ideal resolution. Some users have considered to migrate to other stuff. e.g. https://tech.readyfor.jp/entry/2021/04/15/170227


package.json
```diff
@@ -10,6 +10,7 @@
     "build": "npm run build:development",
     "build:development": "webpack build --mode development",
     "build:production": "webpack build --mode production",
+    "generate:client": "./virtualenv/bin/python manage.py spectacular > spec.yaml && docker run --rm -v \"${PWD}:/local\" openapitools/openapi-generator-cli generate -i /local/spec.yaml -g typescript-fetch -o /local/frontend/src/apiclient/autogenerated --additional-properties=typescriptThreePlus=true",
```
syucream, line 13: Here's a point on this PR. It generates spec file, then also generate .ts files from that.

userlocalhost (commented):
Thank you for your great work!
Would you please decoupling autogenerated codes from this repository and write description about how to generate this API client code in the README.

Thank you in advance :)

syucream:
@userlocalhost I did it! Could you take another look? 🙏

userlocalhost (approved):
LGTM 👍

PR #359, opened 2022-01-11
Refactored test implementation to increase reusability

userlocalhost:
* Increase reusability of attrTypeValue variable by using DjangoContext


userlocalhost:
@hinashi Thank you very much for adding tests!

PR #361, opened 2022-01-12
Refactored the process of get_available_attrs in Entry

hinashi:
Reduced the number of SQL with select_related and prefetch_related.
The more attributes and attribute values, the more effective.

|  | time | query |
----|----|----
| now | 2,200msec | 120query |
| refactoring/get_available_attrs | 1,900msec | 67query |
| refactoring/check_permission | 400msec | 71query |
| refactoring/check_permission_another | 300msec | 68query |
| both refactoring | 100msec | 15query |

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -2,7 +2,7 @@ from collections.abc import Iterable
 from datetime import datetime, date
 
 from django.db import models
-from django.db.models import Q
+from django.db.models import Q, Prefetch
 from django.core.cache import cache
 from django.conf import settings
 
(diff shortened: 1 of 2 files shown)


entry/models.py
```diff
@@ -1095,7 +1106,7 @@ def get_available_attrs(self, user, permission=ACLType.Readable):
             attrinfo['last_value'] = AttrDefaultValue[entity_attr.type]
 
             # check that attribute exists
-            attr = self.attrs.filter(is_active=True, schema=entity_attr).first()
+            attr = entity_attr.attr_list[0] if entity_attr.attr_list else None
```
userlocalhost, line 1109: How about to re-write as below
`attr = next(iter(entity_attr.attr_list), None)`

userlocalhost: We don't know the cost of instantiate of `iter()`

userlocalhost (approved):
How great solution it is!? Thank you for your great contribution!!

PR #362, opened 2022-01-12
Refactored the process of check permission

hinashi:
Entry and Attribute permission check was implemented without checking is_public and default_permission.

|  | time | query |
----|----|----
| now | 2,200msec | 120query |
| refactoring/get_available_attrs | 1,900msec | 67query |
| refactoring/check_permission | 400msec | 71query |
| refactoring/check_permission_another | 300msec | 68query |
| both refactoring | 100msec | 15query |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 ### Refactored
 * Refactored referral param in advanced search (#326)
+* Refactored the process of check permission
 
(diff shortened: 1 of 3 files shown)


userlocalhost (approved):
Thank you very much for your great work!! It's hard to solve problem, but you did it! Thank you!!

PR #364, opened 2022-01-13
Added the function to customize login view

hinashi:
Because the usage and login method change depending on the environment,
Added the function to customize login view.


userlocalhost (approved):
LGTM 👍 Thank you for your good work!

PR #370, opened 2022-01-22
Remove unused LeftMenu component

syucream:
SSIA as refactoring, with updating some snapshots to pass the test


userlocalhost (approved):
Thank you for fixing a CI failure caused by snapshot tests 👍

PR #371, opened 2022-01-24
Added to be able to insert custom html of head

hinashi:
It's a setting change to use GTM(Google Tag Manager)


userlocalhost (approved):
Thank you for your work 👍

PR #373, opened 2022-01-27
Fixed the value when AttributeValue with is_latest set to True is duplicated

hinashi:
Each attribute has multiple AttributeValue associated with it.
There is only one AttributeValue for is_latest=True.
However, there are cases where is_latest=True is duplicated.
(It may not be reproduced in the current version.)

In consideration of duplicate cases, changed so that id uses the last.


userlocalhost (approved):
Great! Thank you for your hard inspection work!! 👍

PR #381, opened 2022-02-02
Set entry-id for each entry columns in the list entry page

userlocalhost:
This is a measure to be able to identify key element from E2E test


hinashi (approved):
LGTM!

PR #385, opened 2022-02-05
Trim too-long breadcrumbs

syucream:
The width of the breadcrumbs makes scrollbar, caused by too long width. 100% + 10% + 10% = 120% 😢 

image


hinashi (approved):
LGTM!
120%... 😢

userlocalhost (approved):
Good catch!! Thank you for your good work 👍

PR #386, opened 2022-02-07
Fixed some problems related with entry's changing history

### LINKED ISSUE #387: Changing values for groups won't be shown correctly
In the entry's changing history page, group and array_group typed attribute values won't be shown correctly when user pushed show_more button as below.

スクリーンショット 2022-02-06 12 44 17

This is a screenshot of entry's changing history at group_array typed attribute that has concrete value, but those values won't be shown correctly. Similar problem will be happened at group typed attribute.

userlocalhost:
Close: https://github.com/dmm-com/airone/issues/387

This commit has following changes for changing history page of entry

* Fixed problem named_entry type value won't be shown correctly
* Fixed problem group and array_group type value won't be shown correctly when it was shown by "make_element_value_text" method
* Fixed minor coding style that is incorrect indent
* Added property to be able to identify each attribute

### DIFF
--- a/templates/show_entry/js_history.html
+++ b/templates/show_entry/js_history.html
@@ -127,7 +127,6 @@ function make_element_value_text(a_type, attrv) {
            `</div>`;
 
   } else if (a_type == attr_type.group) {
-    console.log("[onix-test(group)] attrv: ", attrv);
     return `<a href='/group/edit/${ attrv.value.id }'>${ attrv.value.name }</a>`;
 
   } else if (a_type == attr_type.date) {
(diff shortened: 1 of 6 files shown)


templates/show_entry/js_history.html
```diff
@@ -117,13 +121,14 @@
...
-             `<div class ='col-3'>${ attrv.value.value }</div>` +
-             `<div class ='col-9'>${ ref_value }</div>` +
+    return `<div class='row'>` +
+             `<div class='col-3 attrv_name'>${ attrv.value.value }</div>` +
+             `<div class='col-9 attrv_referral'>${ ref_value }</div>` +
            `</div>`;
 
   } else if (a_type == attr_type.group) {
-    return `<a href='/group/edit/${ attrv.value.id }'>{{ attrv.value.name }}</a>`;
+    console.log("[onix-test(group)] attrv: ", attrv);
```
userlocalhost: This is unnecessary

hinashi (approved):
LGTM!
After deleting the debug log, merge it.

PR #388, opened 2022-02-07
Added npm script for customview

hinashi:
Added script commands for deploying customview.
I made the Thema a separate file to make App.tsx lighter.
(This is to avoid duplicate updates of the AirOne and Customview.)


syucream (approved):
LGTM 👍

userlocalhost (approved):
It's a great contribution to prevent operation mistake!! 👍

PR #391, opened 2022-02-09
Refactored APIv2 and fixed EntryList

userlocalhost:
This commit fixed EntryList page to list entries as alphabetical order.
Along with this fixing, this commit also refactored Entity and Entry v2
API not to send redundant requests for getting Entries that are related
with specified Entity. This integrates getEntity and getEntries API
handler using DRF for returning them with Entity information at the 
same time. By this change, UI processing don't have to send multiple
requests to get Entity and Entires that are related with it.

### DIFF
--- a/entity/api_v2/serializers.py
+++ b/entity/api_v2/serializers.py
@@ -1,0 +1,47 @@
+from airone.lib.acl import ACLType
+from entity.models import Entity
+from entry.models import Entry
+from entry.settings import CONFIG as CONFIG_ENTRY
+from rest_framework import serializers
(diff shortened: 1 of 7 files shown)


hinashi (commented):
We have to implement following features

* check permission for each Entries and target Entity
  (c.f. https://www.django-rest-framework.org/api-guide/permissions/)
* use pagination feature (c.f. https://www.django-rest-framework.org/api-guide/pagination/)

userlocalhost:
It's an outdated PR.

PR #393, opened 2022-02-10
Changed create entry button in entry list for custom view

hinashi:
- Added an argument for custom entity that prohibits create entry.
- Enabled to specify the render in the custom view.
- Enabled to receive the parameters of the React-Router.


userlocalhost (approved):
Thank you for your work!!

PR #399, opened 2022-02-15
Update entry template and expand response context for E2E test

userlocalhost:
This commit update template of restore entry page and expand internal
API for Entry to be able to identify each attribute from E2E test.


hinashi (approved):
LGTM!

PR #402, opened 2022-02-16
Fixed problem to return attribute information that has already been deleted

### LINKED ISSUE #400: The entryAPI returns attribute information that has already been deleted
Even through an EntityAttr is delete, entryAPI returns it as below.
スクリーンショット 2022-02-16 12 06 17

userlocalhost:
Close: #400


hinashi (approved):
LGTM!

PR #403, opened 2022-02-16
Update to use APIv2 to get Entries that are associated with specific Entity

userlocalhost:
This commit adds an API endpoint to get Entries that are related with specified Entity and change new-ui EntryList to use it.

And previous implementation uses '/entry/api/v1/get_entries' that returns Entries without sorting to list Entries of specific Entity. Instead of that, this commit use '/entry/api/v2/entries' that also returns Entries of associated Entity, sorted by alphabetical order. And that API endpoint plans to apply pagination feature.

### DIFF
--- a/entry/api_v2/serializers.py
+++ b/entry/api_v2/serializers.py
@@ -15,8 +15,8 @@ class GetEntrySerializer(serializers.ModelSerializer):
 
     def get_schema(self, entry) -> Dict[str, Any]:
         return {
-                'id': entry.schema.id,
-                'name': entry.schema.name,
+            'id': entry.schema.id,
+            'name': entry.schema.name,
         }
 
     def get_attrs(self, obj) -> Dict[str, Any]:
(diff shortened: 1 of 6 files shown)


hinashi (commented):
There seems to be a performance problem with a large number of entries.
image

It seems that the following measures are necessary.
Omit attrs.
Add pagination.

Would you like to merge this as another issue?

userlocalhost:
Thank you for your comment. I fixed performance problem by omitting Attributes and its values at the "/entry/api/v2/entries" API endpoint handler. Here is the result of this change, that waiting time reduced about one sixth.

### Previous implementation result
スクリーンショット 2022-02-21 9 51 10

### Fixed implementation result
スクリーンショット 2022-02-21 9 52 42

But I want to fix the feature of adding pagination at another PR because it might affect wide range place.

hinashi (approved):
LGTM!

PR #405, opened 2022-02-19
Show job status and actions appropriately

syucream:
Show job status and actions (rerun, cancel) based on each job data.

image


userlocalhost (approved):
LGTM 👍 Thank you for your work!

PR #412, opened 2022-02-24
Changed to add Entity ID to PATH on entry detail page

hinashi:
There is a case to customize the page for each entity in the custom view
So I added the entity ID to the URL.


userlocalhost (approved):
LGTM 👍

syucream (approved):
LGTM

PR #423, opened 2022-02-27
Use API client v2 for job api

syucream:
SSIA


syucream:
It'll be unnecessary if https://github.com/dmm-com/airone/pull/591 is merged.

PR #424, opened 2022-02-28
Fixed to add Entity ID to PATH on entry detail page

hinashi:
Changed to return the schema of object type attribute in entry APIv2.
But advanced search is not yet supported.
image


userlocalhost (approved):
LGTM 👍

PR #409, opened 2022-02-22
Added feature of pagination for EntryListPage

userlocalhost:
### Rest of work
  - [x] Add test for `/entry/api/v2/entries/<entity_id>` endpoint handler associated with adding pagination feature

### Future work
* search function to narrow down entries

PR #427, opened 2022-03-01
Added feature to be able to narrow down entries at EntryList

userlocalhost:
This implemented feature, which is rest of https://github.com/dmm-com/airone/pull/409, to narrow down entries by user specified keyword.

スクリーンショット 2022-03-01 15 47 34


hinashi (approved):
LGTM!

PR #428, opened 2022-03-01
Fixed openapi-generator version to 5.4.0

hinashi:
SSIA


userlocalhost (approved):
Thank you for your commit. I understood that the latest version of `openapi-generator` is `v6.0.0`, through the `v5.4.0` is the latest stable version as below.
(c.f. https://hub.docker.com/r/openapitools/openapi-generator-cli/tags)

PR #436, opened 2022-03-03
Fixed smooth scroll in entry details page

hinashi:
Added react-scroll library.
https://github.com/fisshy/react-scroll

https://user-images.githubusercontent.com/5561786/156500783-71573abc-8eec-446b-8e24-6e9d0fabfcbf.mp4


userlocalhost (approved):
Thank you for your work! It's an elegant solution 👍

PR #438, opened 2022-03-04
Reverted the initialization part of Elasticsearch class

hinashi:
The environment variables and class variables at the time of ESS class initialization were different.
Because to override environment variables during testing.
https://github.com/dmm-com/airone/blob/f511a96345baa43f6d4a1034c47845b73cc61da0/airone/lib/test.py#L14-L20
image


userlocalhost (approved):
Thank you for your work!!

PR #442, opened 2022-03-08
Fixed problem that none of AttributeValue have is_latest is True

### LINKED ISSUE #441: There is a problem that none of AttributeValue have is_latest is true
Create and add a new AttributeValue when editing Entry.
At this time, all is_latest is false.

https://github.com/dmm-com/airone/blob/d615cce8aa9a360d0972adde95c02d4646cbc3da/entry/models.py#L678-L723

There are cases where it becomes a problem if there is no is_latest is True.
https://github.com/dmm-com/airone/blob/d615cce8aa9a360d0972adde95c02d4646cbc3da/entry/models.py#L1086-L1090

If there is no is_latest, there is also a function that looks at the latest AttributeValue.
https://github.com/dmm-com/airone/blob/d615cce8aa9a360d0972adde95c02d4646cbc3da/entry/models.py#L477-L482

hinashi:
close: https://github.com/dmm-com/airone/issues/441

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -543,9 +543,11 @@ class Attribute(ACLBase):
 
         return cloned_attr
 
-    def unset_latest_flag(self):
-        AttributeValue.objects.filter(parent_attr=self,
-                                      is_latest=True).update(is_latest=False)
+    def unset_latest_flag(self, exclude_id=None):
+        exclude = Q()
+        if exclude_id:
+            exclude = Q(id=exclude_id)
+        self.values.filter(is_latest=True).exclude(exclude).update(is_latest=False)
 
     def _validate_value(self, value):
         def _is_group_object(val):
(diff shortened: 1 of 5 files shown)


entry/models.py
```diff
@@ -543,9 +543,11 @@ def clone(self, user, **extra_params):
...
         return cloned_attr
 
-    def unset_latest_flag(self):
-        AttributeValue.objects.filter(parent_attr=self,
-                                      is_latest=True).update(is_latest=False)
+    def unset_latest_flag(self, exclude_id=None):
+        exclude = Q()
+        if exclude_id:
+            exclude = Q(id=exclude_id)
+        self.values.filter(is_latest=True).exclude(exclude).update(is_latest=False)
```
userlocalhost, line 550: 👍

userlocalhost (approved):
Nice work!!

PR #445, opened 2022-03-09
Expanded the scope of mypy

hinashi:
Expanded the scope of mypy from `airone/lib` to everything.
Also, Corrected the pointed out point.


userlocalhost (approved):
Thank you for making this good work!

PR #446, opened 2022-03-10
Fixed that can be retrieved without permission in Entry API v2

hinashi:
Added process to check permissions in EntryAPIv2

image


userlocalhost (approved):
Thank you very much for fixing very meaningful work!!

syucream (approved):
LGTM

Issue #440, opened 2022-03-07
(React / EntryDetails) ignores Attribute's index


hinashi:
There is a problem that the order of attributes is different from index.
It is similar to the following problem.
https://github.com/dmm-com/airone/issues/429

- now
![image](https://user-images.githubusercontent.com/5561786/156958820-a04b5538-12c8-43fa-b4a4-92fa55146e42.png)

- new
![image](https://user-images.githubusercontent.com/5561786/156958845-b89cc82c-69fe-4f49-99e0-9f277b1e6fd7.png)

Issue #435, opened 2022-03-02
(React / EntryDetails) AttributeValue without Attribute


hinashi:
There is a problem that AttributeValue without Attribute is not displayed.

The reproduction procedure is as follows.
- Create Entity
- Create Entry
- Add Attribute to Entity
- View EntryDetails

![image](https://user-images.githubusercontent.com/5561786/156299389-ee456040-84b9-4665-8620-5faa6264473a.png)
![image](https://user-images.githubusercontent.com/5561786/156299412-d1a84528-a4d9-4b00-9061-b5310f4f7f16.png)

PR #447, opened 2022-03-12
Resolve entry attrs based on entity attrs

syucream:
To fix:

- https://github.com/dmm-com/airone/issues/435
- https://github.com/dmm-com/airone/issues/440

I dit:

- Sort attributes by index
- Return list of attributes instead of dict on the API V2 to keep the order

As a result:

image
image


hinashi (approved):
LGTM!

userlocalhost (approved):
LGTM 👍 Thank you for your good work!!

Issue #429, opened 2022-03-02
API response ignores Attribute's index


userlocalhost:
When user access entry detail's page, the attribute values are ordered by `index` value that is set at corresponding `EntityAttr`. But, when user get contents from API `/api/v1/entry`, the order of attribute value isn't cared.

These are the results of Entry's data via Web page and API response. The `Entity01` has three attributes and the order (index value) of these attributes are following.
| Attribute Name | Index |
|----------------|------|
| attr3 | 1 |
| attr2 | 2 |
| attr1 | 3 |

<img width="601" alt="スクリーンショット 2022-03-02 10 59 44" src="https://user-images.githubusercontent.com/469934/156280807-8546e530-1753-47e6-8072-91ad4f1ae030.png">

The context of API response ignores Attribute's index (it returns attribute data with arbitrary order) as below.

<img width="367" alt="スクリーンショット 2022-03-02 10 59 38" src="https://user-images.githubusercontent.com/469934/156280825-62cf86db-c04e-4eb4-92fc-7fb74d66abad.png">

PR #448, opened 2022-03-13
Sort attrs by index on entry api v1

syucream:
https://github.com/dmm-com/airone/issues/429

Use `index` to sort entry attributes on Entry API v1. As a result:
image
image

Btw I wonder if we should return attributes not stored in Attribute but stored in EntityAttribute, like `attr5` in the image 🤔

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -1190,7 +1190,15 @@ class Entry(ACLBase):
                 not user.has_permission(self, ACLType.Readable)):
             return None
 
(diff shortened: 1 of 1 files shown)


hinashi (approved):
LGTM!
It more better to return attributes not stored in Attribute but stored in EntityAttribute.

userlocalhost (approved):
It's great to catch up with latest Issue 👍

PR #450, opened 2022-03-15
Changed to separate settings for each environment by django-configrations

hinashi:
Changed to separate settings for each environment by django-configrations.

(c.f. https://django-configurations.readthedocs.io/en/stable/)

- for prd env
`airone.settings.Prd` write override setting
export DJANGO_CONFIGURATION=Prd


userlocalhost (approved):
It's a brilliant solution!! Thank you very much!!

Issue #430, opened 2022-03-02
(React / EntryList) Behavior when the export button is pressed


hinashi:
There is a problem I don't know if it's running because nothing happens after press the export button.

![image](https://user-images.githubusercontent.com/5561786/156287921-ecbd6d92-d8e9-4119-ab6c-d0a53247d549.png)

Issue #416, opened 2022-02-25
(React / EntityList) show message on delete error


hinashi:
Deleting an entity will result in an error if there are any remaining entries.
![image](https://user-images.githubusercontent.com/5561786/155658499-3ccaf254-8a88-4844-a4db-cb197c5a5cb4.png)

PR #453, opened 2022-03-19
Separate EntityControlMenu as a common component

### LINKED ISSUE #431: (React / EntityList) Entity operation menu item
Both the entity list and the entry list have entity operation menus, but the items are different.
I think it's better to make them the same.

- EntityList
image

- EntryList
image

syucream:
To resolve https://github.com/dmm-com/airone/issues/431, and experimentally implement https://github.com/dmm-com/airone/issues/416 and https://github.com/dmm-com/airone/issues/430

It has common component on entity list page and entry list page.

- entity list page
image

- entry list page
image
 
with notifying asynchronous events on Snackbar for experiment.
image


frontend/src/App.tsx
```diff
@@ -11,7 +12,10 @@ const App: FC = () => {
   return (
     <ThemeProvider theme={theme}>
       <ErrorHandler>
-        <AppRouter />
+        {/* TODO does it unable to apply custom styles? reimplement it ourselves? */}
```
syucream, line 15: It depends on https://github.com/iamhosseindhv/notistack to make easy to notify events on Snackbar entirely. But, unfortunately it seems not to support custom styles with MUI-v5. https://smartdevpreneur.com/customizing-notistack-with-material-ui-snackbar/

I'm considering if its better to reimplement similar stack ourselves or resolve the problem with notistack 🤔

hinashi: So easy to make and great!
I think the design is no problem for now.

hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for your great work!!

PR #455, opened 2022-03-22
Changed job result backend of celery from rabbitmq to None

hinashi:
We need to monitor the number of queues in rabbitmq to check the health of job passing from django to celery.
At that time, if save the job result of celery in rabbitmq, it will interfere with monitoring.


userlocalhost (approved):
LGTM! 👍

PR #456, opened 2022-03-22
Fixed duplicate to send email at the time of error

hinashi:
Disable the default mail_admins handler for django.

disable_existing_loggers
(c.f. https://docs.djangoproject.com/en/4.0/topics/logging/#configuring-logging-1)


userlocalhost (approved):
Thank you for your work!!

PR #457, opened 2022-03-22
Add event-handler onKeyPress at SearchBox, but it has problem...

userlocalhost:
This PR adds event handler onKeyPress at SearchBox, but it fails test because auto-generated snapshot removes SearchBox (maybe because of this adding event-handler)


syucream (approved):
LGTM

hinashi (approved):
LGTM!

PR #458, opened 2022-03-22
Added add and remove attributes depending on entity in get entry api v2

hinashi:
Because I want to add an attribute to the entity in the custom view.


userlocalhost (approved):
LGTM 👍

PR #459, opened 2022-03-22
Changed entry copy page url for customview

hinashi:
Added entity ID to URL to add custom view to entry copy page


userlocalhost (approved):
LGTM 👍

Issue #408, opened 2022-02-22
An exception would be raised at Celery worker when user specify very long attribute name at Entity


userlocalhost:
SSIA

PR #461, opened 2022-03-26
Validate length of entity/attr name

syucream:
https://github.com/dmm-com/airone/issues/408

It accepts submitted data contains too long attribute name, but Celery worker causes errors on writing it to DB. So this PR validates the length before.

### DIFF
--- a/entity/views.py
+++ b/entity/views.py
@@ -101,12 +101,15 @@ def edit(request, entity_id):
 
 
 @http_post([
-    {'name': 'name', 'type': str, 'checker': lambda x: x['name']},
+    {'name': 'name', 'type': str, 'checker': lambda x: (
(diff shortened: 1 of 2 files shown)


entity/views.py
```diff
@@ -101,12 +101,15 @@ def edit(request, entity_id):
 
 
 @http_post([
-    {'name': 'name', 'type': str, 'checker': lambda x: x['name']},
+    {'name': 'name', 'type': str, 'checker': lambda x: (
+            x['name'] and len(x['name']) <= Entity._meta.get_field('name').max_length
```
hinashi, line 105: > Entity._meta.get_field('name').max_length

I didn't know about there is meta infomation!

hinashi (approved):
LGTM!

userlocalhost (approved):
How excellent to solve this problem. Thank you!!

PR #466, opened 2022-03-29
Fixed request even if the password is empty on change ldap auth

### LINKED ISSUE #465: Request even if the password is empty on change ldap auth
If the password is empty, verify that it will not be requested.
image

hinashi:
close: https://github.com/dmm-com/airone/issues/465

image


userlocalhost (approved):
Thank you for your constant good job 👍

PR #468, opened 2022-03-30
Fixed import failing when testing with extensions

hinashi:
I changed the part to import because the extension test fails.

This is due to the following changes.
https://github.com/dmm-com/airone/pull/450/files#diff-7107a7b2d125d397382a3cc56d2cee11856b7dc9dca7da547319d912b16a31bc


userlocalhost (approved):
Thank you for fixing it!

PR #472, opened 2022-04-01
Added custom processing when after delete entry

hinashi:
This is because you may want to perform processing after deleting an entry.


userlocalhost (approved):
LGTM 👍  Thank you for your work!

Issue #469, opened 2022-03-30
The snack bar disappears soon


hinashi:
Deleting an entity with an entry present will result in an error.
The error snack bar disappears shortly after it appears.

https://user-images.githubusercontent.com/5561786/160787675-3ec8ded6-2135-4303-aa83-bba6781eecf2.mp4

PR #473, opened 2022-04-02
Don't notify if routing

syucream:
Just a mitigation for https://github.com/dmm-com/airone/issues/469

- Don't perform routing if it fails, just notify it
- Don't show success message if it doesn't fail

because the notification and routing conflict.


hinashi (approved):
LGTM
> because the notification and routing conflict.

I got it.

userlocalhost (approved):
Thank you for your work!

PR #474, opened 2022-04-02
Fuzzy match filter keywords

syucream:
Allow to match full-size text with search targets like entity name, by converting it implicitly.


frontend/src/utils/StringUtil.ts
```diff
@@ -0,0 +1,12 @@
+function toHalfWidth(origin: string): string {
```
syucream, line 6: ref. https://www.yoheim.net/blog.php?q=20191101

hinashi (approved):
LGTM! this is easy to use.

userlocalhost (approved):
Thank you very much for your excellent work 👍 I like it!

If you never mind, I want to use explanatory constant than raw one. Furthermore, I added description what it is because this must be a genius's work, which is hard to understand later for normal people :D

PR #475, opened 2022-04-13
Change to set ACL to Role instance instead of Users/Groups

userlocalhost:
This commit changed implementation to be able to set ACL by Web UI


hinashi (approved):
LGTM!

PR #476, opened 2022-04-15
Fixed gunicorn and celery failing to start when use extensions

hinashi:
Fixed following PR changes that caused gunicorn and celery to fail to start.
https://github.com/dmm-com/airone/pull/468


userlocalhost (approved):
LGTM. Thank you for checking problems in detail 👍

Issue #413, opened 2022-02-25
(React / EntityList) Display is slow


hinashi:
Display is slow when there are many entities.
![image](https://user-images.githubusercontent.com/5561786/155647381-0758725b-901d-4621-9061-cc489f352835.png)

PR #477, opened 2022-04-16
Make entity list page faster

syucream:
https://github.com/dmm-com/airone/issues/413

by performing pagination on server side.


frontend/src/pages/DashboardPage.tsx
```diff
@@ -43,8 +43,9 @@ export const DashboardPage: FC = () => {
   const [entryQuery, setEntryQuery] = useState("");
 
   const entities = useAsync(async () => {
-    const entities = await aironeApiClientV2.getEntities();
-    return entities.filter((e) => e.isToplevel);
+    // TODO get only top level entities with server-side filters
+    const entities = await aironeApiClientV2.getEntities(1);
```
syucream, line 47: If it gets all entities from API, then filter them by isTopLevel on client side, the operation is so heavy.

Ideally I perform the filter on server side. How can we do it? *isTopLevel* flag is actually a one of the bit flag on status column so we possibly need to write complicated filter condition on Django model ...?

frontend/src/apiclient/AironeApiClientV2.ts
```diff
@@ -74,8 +77,23 @@ class AironeApiClientV2 {
...
+  ): Promise<PaginatedEntityList> {
+    const params = page
+      ? {
+          offset: (page - 1) * ConstEntityList.MAX_ROW_COUNT,
+          limit: ConstEntityList.MAX_ROW_COUNT,
+          query: query,
+        }
+      : {
+          // Any better way to get all the entities?
+          limit: Number.MAX_SAFE_INTEGER,
```
syucream, line 92: Any better way? limit-offset pagination forces us to specify valid limit param.

userlocalhost (approved):
LGTM 👍 Thank you for your work!

hinashi (approved):
LGTM!

PR #478, opened 2022-04-16
Install @types/react-router-dom

syucream:
To get type info supports

### DIFF
--- a/frontend/src/hooks/useTypedParams.tsx
+++ b/frontend/src/hooks/useTypedParams.tsx
@@ -1,0 +1,7 @@
+import { useParams } from "react-router-dom";
+
+export const useTypedParams = <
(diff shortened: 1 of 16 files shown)


frontend/src/hooks/useTypedParams.tsx
```diff
@@ -0,0 +1,7 @@
+import { useParams } from "react-router-dom";
+
+export const useTypedParams = <
+  Params extends { [K in keyof Params]?: any }
```
syucream, line 4: It handles the types of values as `any` so its not type safe, but its equivalent to the current implementation.

userlocalhost (approved):
LGTM! Thank you for your hard work that source code will be more safer 👍

hinashi (approved):
LGTM!

PR #483, opened 2022-04-23
Deny duplicated active entity attribute names

syucream:
Duplicated entity attribute names confuse us, especially narrow down parts.

Ideally I want to implement with UNIQUE constraint in DB, but it requires to ensure uniqueness on the pair of `ACLBase.name` and `EntityAttr.parent_entity_id` so it was impossible. Alternatively I introduce duplication check on Django view.


userlocalhost (approved):
LGTM 👍  Thank you very much for completing very variable work!!

hinashi (approved):
LGTM!

PR #485, opened 2022-04-23
Auto-format with black

syucream:
Request for comment :)

Auto-format with [black](https://github.com/psf/black), like `npm run fix`.

see also https://hesma2.hatenablog.com/entry/2021/01/18/080000
And also Django uses it https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/


userlocalhost:
Thank you for great work! We think to make measure to solve the pep8 error.
* Ignore E203 error
* Ignore E501 error on the "airone/tests/test_elasticsearch.py" file.

hinashi (approved):
LGTM!

PR #486, opened 2022-05-01
Enable to import entries belonging to multiple entities

syucream:
It enables to import entries belonging to multiple entities that we can get with exporting advanced search result, on 1 import job.
For now, we can submit an import job on advanced search result page.

image


templates/advanced_search/result.html
```diff
@@ -25,6 +25,7 @@ <h4>
           </button>
           <button type="button" class="btn btn-primary btn-sm export_yaml">YAML 出力</button>
           <button type="button" class="btn btn-primary btn-sm export_csv">CSV 出力</button>
+          <a class="btn btn-secondary btn-sm" href="/dashboard/import_search_result/">インポート</a>
```
userlocalhost: You may want to put the entry point of this feature at existing place.
User might be used to do this operation at ordinary import way from Entry list page.

userlocalhost (commented):
Thank you very much for implementing feature that users desire to!
But, I'm not sure whether we should make another new UI entry-point for importing. Because there are many ways to import information (from the Entity list page and individual Entry list pages.
Anyway, this must be very useful feature. We are very much looking forward to merging this into this application.
Thank you 👍

templates/advanced_search/result.html
userlocalhost: I changed my mind. It's OK 👍

syucream:
@userlocalhost to simplify importing features, I've unified importing multiple entities on entry import task. How about this?

hinashi (approved):
LGTM!

userlocalhost (approved):
LGTM 👍 Thank you for your hard work!! It's a great job!!

userlocalhost:
Would you please adding line that describes this feature at the CHANGELOG?

PR #489, opened 2022-05-04
Typed entry values

syucream:
entry attribute value has actually some kinds of types, like `string`, `object`, ... For now frontend code specify `any` to skip strict typing. But with API schema, now we specify actual types smoothly, maybe.


frontend/src/apiclient/autogenerated/models/EntryRetrieveValue.ts
```diff
@@ -0,0 +1,166 @@
...
+   * @type {Array<string>}
+   * @memberof EntryRetrieveValue
+   */
+  asArrayString?: Array<string>;
+  /**
+   *
+   * @type {Array<{ [key: string]: object; }>}
+   * @memberof EntryRetrieveValue
+   */
+  asArrayNamedObject?: Array<{ [key: string]: object }>;
```
syucream, line 70: I've specified concrete type in DRF side, but its rounded as `object` ...

syucream, line 70: I've specified concrete type in DRF side, but its rounded as `object` ...

userlocalhost (approved):
Thank you very much for your hard work! It must increase our code safety 👍

PR #492, opened 2022-05-16
Fixed some model and UI problems related with new added Role feature

userlocalhost:
This PR has following fixes for Role feature

* Added `delete()` method that override Django model not to delete it from DB.
* Changed specification at the edit page of Role to list selected them at the top of its list to understand what Role(s) have already been selected.
* Separate create and edit page to distinguish them from UI.


hinashi:
@userlocalhost 
The role edit page was still the create page.
image

hinashi (approved):
LGTM!

PR #493, opened 2022-05-17
Fixed problems associated with setting ACL

userlocalhost:
This commit fixed following problems about ACL

1. A problem that user can set an ACL that no one can access it.
2. A problem that user can change Role's permission that user isn't
   belonged to it as an administrative one.

### DIFF
--- a/acl/views.py
+++ b/acl/views.py
@@ -50,6 +50,7 @@ def index(request, obj_id):
                 "current_permission": get_current_permission(x),
             }
             for x in Role.objects.filter(is_active=True)
+            if x.is_editable(request.user)
         ],
     }
     return render(request, "edit_acl.html", context)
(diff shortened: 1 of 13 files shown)


acl/views.py
```diff
@@ -112,8 +131,10 @@ def set(request, recv_data):
         role = Role.objects.get(id=acl_data["role_id"])
         acl_type = [x for x in ACLType.all() if x == int(acl_data["value"])][0]
 
-        # update permissios for the target ACLBased object
-        _set_permission(role, acl_obj, acl_type)
+        # Role can be cahnged by user who belonged to administrative members
+        if role.is_editable(request.user):
```
hinashi: is_editable -> is_belonged_to ?

(This check may not be needed as it is validated.)

hinashi (approved):
I tried it in a local environment and it seemed to be okay.
Thank you for the fix!

PR #494, opened 2022-05-18
Fixed an error when copying a lot of entries

hinashi:
There is a problem that an error occurs due to the high load of DB when copying a lot of entries.
One job is created for each destination entry.
Since it is a dependent job, it queries the DB once per second.

In this PR, the copy jobs have been combined into one.


userlocalhost (approved):
LGTM 👍

PR #497, opened 2022-05-23
Fixed HTTP 404 exception error

hinashi:
There is a problem that the HTTP404 error becomes an internal server error.
Fixed by changing the order of WhiteNoiseMiddleware.
(c.f. http://whitenoise.evans.io/en/stable/django.html#enable-whitenoise)

Also changed to display an error when the debug parameter is True.

### DIFF
--- a/airone/settings_common.py
+++ b/airone/settings_common.py
@@ -65,15 +65,15 @@ class Common(Configuration):
     ]
 
     MIDDLEWARE = [
-        "airone.lib.log.LoggingRequestMiddleware",
         "django.middleware.security.SecurityMiddleware",
+        "whitenoise.middleware.WhiteNoiseMiddleware",
+        "airone.lib.log.LoggingRequestMiddleware",
         "django.contrib.sessions.middleware.SessionMiddleware",
         "django.middleware.common.CommonMiddleware",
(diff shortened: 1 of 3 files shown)


airone/settings_common.py
```diff
@@ -65,15 +65,15 @@ class Common(Configuration):
     ]
 
     MIDDLEWARE = [
-        "airone.lib.log.LoggingRequestMiddleware",
         "django.middleware.security.SecurityMiddleware",
+        "whitenoise.middleware.WhiteNoiseMiddleware",
```
userlocalhost, line 69: static 以下の静的ファイルを返すモジュール

airone/settings_common.py
```diff
@@ -65,15 +65,15 @@ class Common(Configuration):
...
+        "airone.lib.log.LoggingRequestMiddleware",
```
userlocalhost, line 70: 例外エラーが発生した際にメール送信するモジュール

userlocalhost (approved):
LGTM 👍 How beautiful to solve this problem 👍 Thank you very much!!

syucream (approved):
LGTM

PR #504, opened 2022-06-07
Fixed wrong modifications of some Models default attribute values

userlocalhost:
I changed configurations of default attribute values to "DEFAULT NULL"
from "NOT NULL" on some Models by mistake associated with upgrading
Django from v1.11 to v2.0. This commit fixed it.


hinashi (approved):
LGTM!

PR #507, opened 2022-06-09
Fix broken entry referral API V2

syucream:
> TypeError: object of type 'NoneType' has no len()

Fix this issue


hinashi (approved):
LGTM!

PR #508, opened 2022-06-13
Removed webhook apiv2 in new-ui

hinashi:
Removed webhook because added it to entity apiv2.


userlocalhost (approved):
It's a good work! Thank you for your work 👍

PR #509, opened 2022-06-13
Added entry copy apiv2 in new-ui

hinashi:
SSIA


userlocalhost (approved):
Thank you for your hard work!!

PR #512, opened 2022-06-14
Fixed empty display of array type in advanced search result

hinashi:
If the value is empty, it will not be output in advanced search results.
image


userlocalhost (approved):
LGTM 👍

PR #517, opened 2022-06-17
Added warning dialog in edit entity page

hinashi:
Added a dialog that appears when you leave the entity edit page.
(c.f. https://v5.reactrouter.com/core/api/Prompt)
image


userlocalhost (approved):
Thank you for impelementing this feature by elegant way 👍

PR #520, opened 2022-06-21
Fixed test to be run by Django testing framework

userlocalhost:
The previous method names of this test are not followed by naming
convention of unittest. So those tests didn't run by Django testing
framework. This commit fixed it.


hinashi (approved):
LGTM!

PR #521, opened 2022-06-24
Fixed pip installation failing

hinashi:
Added library installation because python3-SAML installation fails.

When starting AirOne under a proxy such as nginx, the SSO response URL has a different port number and an error occurs.
The USE_X_FORWARDED_PORT option has been enabled and resolved.


userlocalhost (approved):
Thank you for your prompt fixing for a problem 👍

PR #523, opened 2022-06-26
Enable to restore deleted entries

syucream:
Add restore-entry page to support restoring deleted entries on the new UI.

image

```
ALTER TABLE `acl_aclbase` ADD COLUMN `deleted_time` datetime(6) NULL;
ALTER TABLE `acl_aclbase` ADD COLUMN `deleted_user_id` integer NULL , ADD CONSTRAINT `acl_aclbase_deleted_user_id_0a1d3f53_fk_user_user_id` FOREIGN KEY (`deleted_user_id`) REFERENCES `user_user`(`id`);
```


syucream:
Why the snapshot test fails ... 🤔

syucream:
Finally fixed! The snapshots on the CI and on my local env are different. The CI uses Node.js 17.x but my local env used 18.x. I use 17.x then the problem is fixed.

To support Node.js 18.x, we need to prepare some configs like package.json and to fix the snapshots. I 
 create an issue for that https://github.com/dmm-com/airone/issues/526

hinashi:
@syucream 
Thank you for implementing!
Would you please create that dialog will be displayed when click the entry.
image

syucream:
@hinashi I did that!

image

acl/models.py
```diff
@@ -38,6 +39,10 @@ class ACLBase(models.Model):
     status = models.IntegerField(default=0)
     default_permission = models.IntegerField(default=ACLType.Nothing().id)
     updated_time = models.DateTimeField(auto_now=True)
+    deleted_user = models.ForeignKey(
```
syucream, line 42: To record who delete it, when deleted newly.

entry/api_v2/views.py
```diff
@@ -65,7 +65,7 @@ def destroy(self, request, pk):
         user.seth_entry_del(entry)
 
         # delete entry
-        entry.delete()
+        entry.delete(deleted_user=user)
```
syucream, line 68: For now, deleted_user is only recorded at here. But its usable on other ACLBase subclasses.

hinashi (approved):
LGTM!

PR #524, opened 2022-06-27
Fixed webhook some problems

hinashi:
* Fixed the header of webhook is not set
* Fixed entry recovery and attribute value revert not sending webhooks


userlocalhost (approved):
Thank you very much for your great work!! 👍

PR #529, opened 2022-07-05
Upgrade flower version from v1.0.0 to v1.1.0

hinashi:
fixed: https://github.com/dmm-com/airone/security/dependabot/22

Also, requirements and requirements-dev were double-managed, so fixed it.


userlocalhost (approved):
Thank you for doing this meaningful work 👍

PR #532, opened 2022-07-06
Added helper method get_referred_entries at Group instance

userlocalhost:
This commit add a helper method Group.get_referred_entries() that
returns Entry instances which specified Group instance is referred.


hinashi (approved):
LGTM!

PR #539, opened 2022-07-17
Not to show prompt when succeeded to submit

syucream:
If creating/updating an entity is succeeded, no need to show the prompt


userlocalhost (approved):
It's a good work!! Thank you for your prompt work 👍

hinashi (approved):
LGTM!

PR #540, opened 2022-07-20
Added npm package for use customview

hinashi:
Added the npm package for use customview.
It would be nice to be able to split for customview...


userlocalhost (approved):
LGTM 👍

PR #550, opened 2022-07-25
Fixed custom_view not applied when importing multiple entities

hinashi:
Changed to call custom view for each entity.


userlocalhost (approved):
Thank you very much for fixing critical problems. It must make user happy 👍

PR #553, opened 2022-07-30
Reload the entity list after deleting an entity

syucream:
Fix not to show a deleted entity on the entity list page.

### DIFF
--- a/frontend/src/components/entity/EntityControlMenu.tsx
+++ b/frontend/src/components/entity/EntityControlMenu.tsx
@@ -17,6 +17,7 @@ import {
   entitiesPath,
   importEntriesPath,
   restoreEntryPath,
+  topPath,
 } from "Routes";
 import { aironeApiClientV2 } from "apiclient/AironeApiClientV2";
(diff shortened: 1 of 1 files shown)


frontend/src/components/entity/EntityControlMenu.tsx
```diff
@@ -45,6 +46,8 @@ export const EntityControlMenu: FC<Props> = ({
         enqueueSnackbar("エンティティの削除が完了しました", {
           variant: "success",
         });
+        // A magic to reload the entity list with keeping snackbar
+        history.replace(topPath());
```
syucream, line 50: If a new path is same with current page, react router doesn't re-render the current page (so the page can show a deleted entity).

To avoid it, it switches to another path for a moment, then re-switch to the path to re-render. As an another way, we can call `history.go(0)` to reload, but it will hide a snackbar.

userlocalhost (approved):
Thank you for good work!

hinashi (approved):
LGTM!

PR #555, opened 2022-07-30
Redirect to restore-entry page if accessing to entry details page has been deleted

syucream:
Like the current UI. To avoid to operate to a deleted entry accidentally.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for making important work with an elegant code 👍

PR #558, opened 2022-08-02
Update job model to introduce new status "WARNING"

userlocalhost:
This commit introduced new Job's status WARNING, that is
intermediate status between succeed and error.


userlocalhost (commented):
We need to update template processing.

hinashi (approved):
LGTM!

PR #559, opened 2022-08-04
Fixed error in entry details page new ui

hinashi:
In entry APIv2, the value changes for each type of attribute.
(e.g. `value: {as_string: "hoge"}`)

Added sending is_v2 variable to custom view.


userlocalhost (approved):
LGTM 👍

PR #562, opened 2022-08-06
Implement entry attribute referrals API V2

syucream:
Replace `get_attr_referrals`  with a new endpoint in API V2.


entry/api_v2/views.py
```diff
@@ -359,3 +362,40 @@ def post(self, request, entity_id: int):
...
+        else:
+            entity_attr = EntityAttr.objects.filter(id=attr_id).first()
+        if not entity_attr:
+            raise NotFound(f"not found matched attribute or entity attr: {attr_id}")
+
+        conditions = {"is_active": True}
+        if keyword:
+            conditions["name__icontains"] = keyword
+
+        # TODO support natural sort?
```
syucream, line 391: In API V1, it performs `natsort`. its not replacable in queryset ... is it required? should we perform it on the client side?

hinashi (approved):
LGTM!

entry/api_v2/views.py
userlocalhost: It's OK to implement with your work. Let's fix it when any problem will be emerged 👍

userlocalhost (approved):
LGTM 👍 Thank you for your hard work!!

PR #565, opened 2022-08-08
Refactored APIv2 for custom view

hinashi:
Defined the type of data to send to the custom view.
Renamed custom views in v1 and v2.


userlocalhost (approved):
Thank you very much for making more convenient to implement custom-view feature 👍

PR #567, opened 2022-08-12
Fixed to parse celery exception error log

hinashi:
Normally, application logs are sent to a log server and aggregated.
At that time, if there is a long message such as an exception error, 
there is a problem that the format is different from other logs and cannot be parsed.
Added a simple log that can be parsed when an exception occurs in celery.

e.g.
`[ERROR] asctime:2022-08-12 10:11:45,867 module:celery   message:An exception error has occurred process:10756   thread:140528500768896`

### DIFF
--- a/airone/celery.py
+++ b/airone/celery.py
@@ -55,4 +55,9 @@ full traceback:
 
     # Logger for DEBUG because email is not sent in dev environment
     Logger.error(message)
+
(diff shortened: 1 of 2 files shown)


syucream (approved):
LGTM

userlocalhost (approved):
LGTM 👍 Thank you very much to measure Datadog's log parser problem.

PR #568, opened 2022-08-12
Changed to update job status to error when celery exception error

hinashi:
Jobs have a mechanism of dependent jobs.
When an exception error occurred in celery, there was a problem that dependent jobs were stuck.
Resolved by changing job status to update to error.


syucream (approved):
LGTM

userlocalhost (approved):
It's a very good work!! LGTM 👍

PR #570, opened 2022-08-13
Switch to entry list page on canceling to create an entry

syucream:
SSIA, not to fail to show entry detail page. The entry detail page is unavailable for uncreated entry :)


userlocalhost (approved):
LGTM 👍

hinashi (approved):
LGTM!

PR #571, opened 2022-08-13
Replace Fab with Button

syucream:
It replaces Fab as a create-button with Button to avoid forcefully show the button on overlapping other components. 
image

The Fab has strong z-index value (1050) by default because it's just floating action button. I think is not suitable for create-button so I replaced it.
https://mui.com/material-ui/customization/z-index/


hinashi (approved):
LGTM!

userlocalhost (approved):
Good catch 👍 Thank you for fixing it!!

PR #572, opened 2022-08-13
Adjust anchor link position with considering fixed header etc.

syucream:
Currently anchor links don't work correctly because of fixed components ... 😭 

https://user-images.githubusercontent.com/191684/184480334-2baa2948-a1e5-463a-9ea4-1c12362c945f.mov


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for fixing that problem with this elegant code 👍

PR #573, opened 2022-08-15
Update processing to prevent creating unnecessary NOTIFY_UPDATE_ENTRY Job at EntryAPI of api.v1

userlocalhost:
In the previous implementation, EntryAPI.post create NOTIFY_UPDATE_ENTRY Job whenever it is called.
This commit refactored it that only create its Job when target Entry attribute values or its name are actually updated.


hinashi (approved):
LGTM!

PR #575, opened 2022-08-17
Changed not to notify when no update in entry update

hinashi:
SSIA


userlocalhost (approved):
It's a good work to prevent to sending unnecessary request to AirOne 👍

PR #576, opened 2022-08-18
Changed to allow parallel execution some job

hinashi:
Some jobs, such as NOTIFY_UPDATE_ENTRY, can safely ignore dependencies.
So I enabled parallel execution of those jobs.


userlocalhost (approved):
LGTM 👍

PR #578, opened 2022-08-19
Fixed logic to disable create button on entry list page

hinashi:
c.f. https://github.com/dmm-com/airone/pull/571
A feature that disables create button on some entities.

- enabled
image
- disabled
image


userlocalhost (approved):
LGTM 👍 Thank you for making an important work!

PR #579, opened 2022-08-19
Added exclude specified attributes on entry edit page

hinashi:
For some entities, you may not want users to be able to update attributes.
Added a function to exclude specific attributes from being edited.


userlocalhost (approved):
LGTM 👍 Thank you for implementing a necessary feature!

PR #580, opened 2022-08-20
Update selected referrals on changing order of attributes

syucream:
before the selected referrals are not updated appropriately when changing order of attributes, so the referrals look like being  broken:
image


userlocalhost (approved):
LGTM 👍

hinashi (approved):
LGTM!

PR #581, opened 2022-08-20
Show error details simply if entry modification failed

syucream:
Show error messages responded from API on the snackbar if entry creation/update failed.

image

IMO, we need to improve it more. The error message from API is not so human readable.


hinashi (approved):
LGTM!

userlocalhost (approved):
LGTM 👍 Thank you for improving Error showing processing. In near future, We want to decouple error message and error code.

PR #582, opened 2022-08-22
Updated README

hinashi:
SSIA


userlocalhost (approved):
It's a great work! We really appreciate for an important work like this!!

PR #583, opened 2022-08-22
Added side content space on entry details page

hinashi:
Changed custom views to allow adding content to the right of entry details.
image


userlocalhost (approved):
LGTM 👍

PR #584, opened 2022-08-22
Added additional contents on entry details page

hinashi:
Added function to display additional contents on entry details page.

image


userlocalhost (approved):
LGTM 👍

PR #585, opened 2022-08-24
Fixed an error in getting data_value in advanced search

hinashi:
In advanced search, there are cases where the display of date type results in an error.
This is when AttributeValue does not exist in the date type Attribute.

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -1543,6 +1543,7 @@ class Entry(ACLBase):
                 "type": entity_attr.type,
                 "key": "",
                 "value": "",
+                "date_value": None,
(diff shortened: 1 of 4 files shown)


entry/models.py
```diff
@@ -1543,6 +1543,7 @@ def _set_attrinfo(entity_attr, attr, attrv, container, is_recursive=False):
                 "type": entity_attr.type,
                 "key": "",
                 "value": "",
+                "date_value": None,
```
userlocalhost, line 1546: 👍

userlocalhost (approved):
LGTM 👍

PR #586, opened 2022-08-24
Added exclude entity parameter

hinashi:
There are cases where you want to exclude some entities with search_entries_for_simple and get_referred_objects.
Added parameter to exclude.


userlocalhost (approved):
LGTM 👍 Thank you for your hard work.

PR #592, opened 2022-08-29
Added processing for safety's sake not to create duplicate Entries in accident

userlocalhost:
This commit adds safety's sake processing just in case for the situation to create
duplicate Entries when multiple requests passed through existence check.
By this processing, Even through multiple requests coming here, Django prevents
from creating multiple Entries.


hinashi (approved):
LGTM!

PR #593, opened 2022-08-30
Added special search character to get Entries that have substantial attribute values

userlocalhost:
This commit added new character(`*`) for search processing to retrieve Entries that have substantial attribute values.

Current implementation has another special search-character (`\`) to retrieve Entries that have empty attribute values. The feature that is added in this commit is opposite of this.

## Note
When you have already registered data to the Elasticsearch, you have to recreate index mapping by this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
   Contributed by @hinashi
 * Added exclude entity parameter in get_referred_objects.
   Contributed by @hinashi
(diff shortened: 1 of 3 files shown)


userlocalhost:
## Remaining work
- [x] Updating CHANGELOG

hinashi (approved):
LGTM! Thanks for the fix.

PR #595, opened 2022-08-31
(WIP) Enable to select Entity to register the Elasticsearch

userlocalhost:
In the current implementation, all Entries will be registered to the Elasticsearch.
This commit added an option to select whether Entry will be register to the Elasticsearch at creating/editing Entity.

## Application Result
スクリーンショット 2022-08-31 17 40 01

### DIFF
--- a/entity/models.py
+++ b/entity/models.py
@@ -49,6 +49,10 @@ class Entity(ACLBase):
     STATUS_CREATING = 1 << 1
     STATUS_EDITING = 1 << 2
 
+    # Entries that are related with this Entity won't be registered to the elasticsearch
+    # when this status is set
+    STATUS_NOT_INDEXED = 1 << 3
+
     note = models.CharField(max_length=200, blank=True)
     attrs = models.ManyToManyField(EntityAttr)
 
--- a/entry/models.py
+++ b/entry/models.py
@@ -1636,6 +1636,10 @@ class Entry(ACLBase):
         return document
 
     def register_es(self, es=None, skip_refresh=False):
+        # do nothing when "is_not_indexed" parameter is set on the related Entity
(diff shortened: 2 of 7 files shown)


userlocalhost:
## Remaining work
- [ ] Updating CHANGELOG

entry/models.py
```diff
@@ -1636,6 +1636,10 @@ def _set_attrinfo(entity_attr, attr, attrv, container, is_recursive=False):
         return document
 
     def register_es(self, es=None, skip_refresh=False):
+        # do nothing when "is_not_indexed" parameter is set on the related Entity
```
hinashi, line 1639: I thought it would be better to leave the advanced search as it is and exclude only the simple search.

entity/models.py
```diff
@@ -49,6 +49,10 @@ class Entity(ACLBase):
     STATUS_CREATING = 1 << 1
     STATUS_EDITING = 1 << 2
 
+    # Entries that are related with this Entity won't be registered to the elasticsearch
+    # when this status is set
+    STATUS_NOT_INDEXED = 1 << 3
```
hinashi, line 54: DRF filtering becomes complicated when using the status value.
I would like to add a new field to the model if possible.

https://github.com/dmm-com/airone/blob/85ec8038016852b32367c6a4d4a722d10d9b8967/entity/api_v2/views.py#L120-L123

An independent boolean field like `is_active` can be filtered by the DRF function.
https://github.com/dmm-com/airone/blob/85ec8038016852b32367c6a4d4a722d10d9b8967/entity/api_v2/views.py#L164

hinashi (commented):
Thank you for the change!
I commented.

userlocalhost:
Close it because of changing policy for solving this problem.

PR #598, opened 2022-09-03
Ensure acl.parent follows the parent type

syucream:
Declare `ACLParentType` for acl.parent to convert property names to FE side naming conversion, like `is_public` -> `isPublic`.
Currently `acl.parent` has `is_public`, not `isPublic` so this condition always evaluated it's `true`.
https://github.com/dmm-com/airone/blob/638964baafc16a49f31218fad4601765f8b009fc/frontend/src/pages/ACLPage.tsx#L130


hinashi (approved):
LGTM!

acl/api_v2/serializers.py
```diff
@@ -12,6 +12,12 @@
 from user.models import User
 
 
+class ACLParentType(TypedDict):
+    id: int
+    name: str
+    is_public: bool
```
userlocalhost, line 18: I've learned it by your favor.

userlocalhost (approved):
Thank you for making more useful to develop FE.

PR #600, opened 2022-09-03
disableScrollLock on Menu

syucream:
Show the scroll bar even if MUI Menu is opened.
image

MUI Menu hides the scroll bar if its opened. And the scroll bar doesn't overlay the page on (maybe, especially) Windows environment, so the page width is slightly flickering on opening the Menu. `disableScrollLock` should resolve it.


hinashi (approved):
LGTM!

userlocalhost (approved):
Good catch!! And thank you for fixing it with elegant code!

PR #614, opened 2022-09-10
Improve recent jobs list on the Header

syucream:
Simplify recent jobs list on the Header, just to show job details w/o any link  (but on only export jobs, it contains a download link).
It also contains these improvements:
- Support `created_after` query on the Jobs API V2 to replace `/api/v1/job/`.
- Check/update the latest date when user checks the recent jobs list to manage badge count.

image


frontend/src/utils/AironeAPIClient.ts
```diff
@@ -278,10 +278,6 @@ export function importGroups(formData: FormData): Promise<Response> {
   });
 }
 
-export function getRecentJobs(): Promise<Response> {
-  return fetch(`/api/v1/job/`);
```
syucream, line 282: We can use API V2 :)

job/api_v2/views.py
```diff
@@ -39,4 +47,7 @@ def get_queryset(self):
             ),
         )
 
+        if created_after:
+            query &= Q(created_at__gte=created_after)
```
syucream, line 51: To support filtering recent jobs.

hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for excellent work!! 👍

PR #616, opened 2022-09-12
update ESS on changed role or group name

Ravie403:


role/models.py
```diff
@@ -89,9 +95,69 @@ def delete(self):
...
+        from airone.lib.elasticsearch import ESS
+
+        es_object = ESS()
+        for entry in [x for x in self.get_referred_entries() if x.id != self.id]:
+            entry.register_es(es=es_object)
+
+        if settings.ES_CONFIG:
+            self.unregister_es()
+
+    def register_es(self, es=None, skip_refresh=False):
```
hinashi: No need to add register_es and unregister_es for group and role.
(Because we are using entry.register_es(), entry.unregister_es())

role/models.py
```diff
@@ -89,9 +95,69 @@ def delete(self):
         )
         self.save(update_fields=["is_active", "name"])
 
+        # avoid circular import
+        from airone.lib.elasticsearch import ESS
+
+        es_object = ESS()
+        for entry in [x for x in self.get_referred_entries() if x.id != self.id]:
```
hinashi: `if x.id != self.id` logic is not needed.
If Entry, was used to exclude itself.

hinashi (commented):
Thanks for the PR!
I have posted 2 comments.

And, Please write a test for role.get_referred_entries()
See below for reference.
https://github.com/dmm-com/airone/blob/447065212cc386548f7c1d83614d2aa134bae04c/group/tests/test_model.py#L57-L116

hinashi (approved):
LGTM!

PR #623, opened 2022-09-14
Fixing the omission of adding implementation of API v2 client for User.

userlocalhost:
This commit adds implementation of API v2 client for User, which is generated by Open API generator.


hinashi (approved):
LGTM!

PR #607, opened 2022-09-07
Added feature to be able to make a relationship between Entry and Role

userlocalhost:
This commit made a feature to make a connection between Entry and Role(s)
via AttributeValue. This added following two Attribute types.
* role
* role_array

The implementation of them are quite similar to the Group's one. The identification
variable or Role will be stored at the "value" parameter of AttributeValue.

## Remaining work
* [x] Implementation for new-UI
* [x] Update CHANGELOG

PR #625, opened 2022-09-15
Fixed oversight of correspondence to implement for new Role Attribute types

userlocalhost:
This commit implements processing to return Attribute value history for Role with correct format. (c.f. #607)


hinashi (approved):
LGTM!
I checked the operation of change history in local env.

PR #628, opened 2022-09-15
Blush-up a fixing of ValueError exception on update

userlocalhost:
This commit blushed up #626 that exchange an invalid value to 0, which indicates there is no Entry refers when an Attribute of array_entry type is updated.
(c.f. https://github.com/dmm-com/airone/pull/626#pullrequestreview-1108479184)


hinashi (approved):
LGTM!

PR #629, opened 2022-09-15
Fixed entity's page paginate

ritsuxis:


hinashi (commented):
Thanks for the fix.
I confirmed that it was fixed in my local environment.

Please add a test case for 400error by referring to the following.
https://github.com/dmm-com/airone/blob/5e6e888f515f105894504a5581ccecd685064a75/entry/tests/test_view.py#L152-L160

hinashi (approved):
LGTM!

PR #631, opened 2022-09-15
Fixed response speed in advanced search result page

hinashi:
There was a problem that the response was slow when there were many cases.
Fixed by changing the number to 100.


userlocalhost (approved):
LGTM 👍 Thank you for this improving.

PR #634, opened 2022-09-16
Changed to separate may_remove_referral method

hinashi:
Separated the method to remove related entries from delete method,
so that it can be called from elsewhere.

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -1159,21 +1159,18 @@ class Attribute(ACLBase):
             if updated_data and self.is_updated(updated_data):
                 self.add_value(user, updated_data, boolean=attrv.boolean)
 
-    # NOTE: Type-Write
-    def delete(self):
-        super(Attribute, self).delete()
-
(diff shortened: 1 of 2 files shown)


entry/models.py
```diff
@@ -1159,21 +1159,18 @@ def add_to_attrv(self, user, referral=None, value="", boolean=False):
...
                 return
 
-            entry = Entry.objects.filter(id=referral.id, is_active=True).first()
+            entry: Entry = Entry.objects.filter(id=referral.id, is_active=True).first()
             if not entry:
                 # the case referred entry is already deleted, do nothing
                 return
 
-            if entry.get_referred_objects().count() > 0:
+            if entry.get_referred_objects().exclude(id=self.parent_entry.id).count() > 0:
```
userlocalhost, line 1173: 👍

userlocalhost (approved):
LGTM 👍 Thank you for fixing important problem!

PR #637, opened 2022-09-21
Added processing to check user belonged to role as member or administrative user

userlocalhost:
This commit adds new parameter "as_member" at the Role.is_belonged_to() method to be able to confirm that specified user is belonged to this role as member. This feature is necessary to distinguish user is belonged to its role as normal member or administrative one.
スクリーンショット 2022-09-21 16 50 01

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## In development
 
 ### Added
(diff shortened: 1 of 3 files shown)


hinashi (approved):
LGTM!

Issue #267, opened 2021-09-30
Replace MemcachedCache to alternative because it has already been deprecated in Django v3.2


userlocalhost:
According to the release note of Django v3.2, the `MemcachedCache`, [which is used in the AirOne as backend library for memcached](https://github.com/dmm-com/airone/blob/master/airone/settings.py#L201), seems to be deprecated.
(c.f. https://docs.djangoproject.com/en/3.2/releases/3.2/#id3)

PR #642, opened 2022-09-26
Changed not to use memcached

hinashi:
Not using memcached reduces operational load.

The below issue will be closed.
https://github.com/dmm-com/airone/issues/267


userlocalhost (approved):
LGTM 👍 Thank you very much for excellent work to solve memcached problem.

PR #646, opened 2022-09-30
Fixed error in updating array_object attribute

hinashi:
There is a problem that an error occurs when None is specified when updating the array_object attribute.


userlocalhost (approved):
LGTM 👍 Thank you for fixing it with reasonable code.

PR #649, opened 2022-10-01
Return null if referral is not active

syucream:
Currently the entry details page shows deleted referral(s). To avoid it, it returns null on the API if referral is not active.


hinashi (approved):
LGTM!

userlocalhost (approved):
LGTM 👍 Thank you for fixing this problem!!

PR #650, opened 2022-10-03
Changed not to execute notify job when there is no update in entry import

hinashi:
Importing an entry will call notify_update_entry even if there are no updates.
Changed not to execute notify job when there is no update in entry import.


userlocalhost (approved):
Thank you very much for fixing very important problem!!

PR #651, opened 2022-10-03
Fixed remove_from_attrv for array_named_entry attribute

hinashi:
When using remove_from_attrv with array_named_entry attribute, only referral was specified.
Changed to specify both reference and value.


userlocalhost (approved):
Thank you very much for blushing up exist function 👍

PR #652, opened 2022-10-04
Added feature to search Entries with chaining conditions

userlocalhost:
The "advanced search" feature only get Entries using conditions of adjacent ones.

This added a new API handler that user can retrieve Entries, which refers an Entry
at specific Attribute and that referred Entry refers another Entry. In that way,
user can specify this repeated (chained) reference condition to get Entry
that refers far away Entry without aware of intermediate ones.

By using this API handler, user can get any Entries whatever Entry structure is at one API call.


hinashi (approved):
LGTM!

PR #653, opened 2022-10-07
Implement role list page and role form page in React

syucream:
SSIA

image

image


hinashi (approved):
LGTM!

When create role, the created role was not found because isActive is False.
Submittabal conditions required all user, group, admin_user, admin_group changes.

userlocalhost (approved):
Excellent. How fast you work!!

PR #655, opened 2022-10-12
Changed Elasticsarch update process to faster

hinashi:
There was a problem that the update process of ElasticSearch was slow.
Improved as below.
- Use prefetch to speed up database reads
- Use bulkAPI to write ElasticSearch
- Process each entity in parallel


userlocalhost (approved):
Excellent!! It's a great work to improve registering data to the Elasticsearch!! 👍

PR #657, opened 2022-10-13
Added processing to be able to get Entries that refers other Entry from elasticsearch

userlocalhost:
## Remaining work
- [x] When an Entity name is updated, changed to run Entry.register_es() for each Entries, which are subordinated by its Entity.
  (c.f. https://github.com/dmm-com/airone/pull/657/commits/88b2f2652ff706e2d033bc73f5b534a01e512ce5)
- [x] When an Entry(e0) is update to refer another Entry(e1), it's needed to call Entry.register_es() not only e0, but also e1 for updating "referrals" parameter of e1.
  (c.f. https://github.com/dmm-com/airone/pull/657/commits/6d2de8c03bb585e5c77974df8d5fa537e7626656)

- [x] When an Entry(e0) is updated as below. It should update es-documents, both current refers Entry(`ref-1`) and previous referred Entry(`ref-0`)
  - AsIs: An attribute "attr" of e0 refers ref-0
  - ToBe: An attribute "attr" of e0 refers ref-1
  (c.f. https://github.com/dmm-com/airone/pull/657/commits/12596b15afbeac8c5a5e1befc11d0b7bedfe63a4)


hinashi (approved):
LGTM!

PR #660, opened 2022-10-18
Upgrade Elasticsearch version from 6.8.16 to 7.17.6

hinashi:
Changes due to Elasticsearch version upgrade are as follows.
- mapping name was removed.
- response format of value count has changed.
- stats result has changed.


userlocalhost (approved):
Great!! Thank you for making a great improvement.

PR #662, opened 2022-10-20
Removed processing that creates unnecessary Attribute instances

userlocalhost:
This is a workaround processing not to create duplicate Attribute instances just after creating Entry by notify event job.


hinashi (approved):
LGTM!

PR #668, opened 2022-10-27
Fixed problem of exported data structure at referral schema

userlocalhost:
This PR fixes problem of YAML exported data structure of advanced search, that referral schema is not compatible with previous version.


hinashi (approved):
LGTM!

PR #669, opened 2022-10-27
Added API handler that searches Entries across multiple referral structure (/api/v1/search_chain)

userlocalhost:
SSIA


hinashi (approved):
LGTM!

PR #670, opened 2022-10-29
Reorganize showing an error page

syucream:
Reorganize errors handled by ErrorBoundary, error pages and errors occurred on an async process.
It shows some well-designed error pages:

image

image


frontend/src/hooks/useAsyncWithThrow.tsx
```diff
@@ -0,0 +1,49 @@
...
+      loading: false;
+      value?: undefined;
+    }
+  | {
+      loading: false;
+      value: T;
+    };
+
+// A thin wrapper of useAsync() in react-use, but it will throw an error
+export const useAsyncWithThrow = <T extends FunctionReturningPromise>(
```
syucream, line 29: `useAsync()` in react-use returns an error by return value. I want throwing an error, then catch it on ErrorBoundary so I create a thin `useAsync()` wrapper throwing an error if occurred instead of including in return value.

group/api_v2/views.py
```diff
@@ -1,15 +1,30 @@
...
 from group.api_v2.serializers import (
     GroupCreateUpdateSerializer,
     GroupSerializer,
     GroupTreeSerializer,
 )
 from group.models import Group
+from user.models import User
+
+
+class UserPermission(BasePermission):
```
syucream, line 13: to allow only admin to edit(create/update/destroy) a group

frontend/src/ErrorHandler.tsx
```diff
@@ -77,21 +83,48 @@ const GenericError: FC<GenericErrorProps> = ({ children }) => {
...
+    case ForbiddenError.errorName:
+      return <ForbiddenErrorPage />;
+    case NotFoundError.errorName:
+      return <NotFoundErrorPage />;
     default:
       return <GenericError>{error.toString()}</GenericError>;
   }
 };
 
+const ErrorBridge: FC = ({ children }) => {
```
syucream, line 95: An error occurred on an async process cannot caught by ErrorBoundary.
ref. https://www.asobou.co.jp/blog/web/error-boundary

To catch it appropriately, I set a `unhandledrejection` handler to throw those errors explicitly.

hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for your hard work!!

PR #671, opened 2022-10-29
Implement import/export feature for users/groups

syucream:
SSIA

The import/export feature is currently mixed with users and groups, but the new feature implemented in this PR is separated.

image

image


hinashi (approved):
LGTM!

("A user with that username already exists.", "group with this name already exists."
error occurred when I imported the exported data.)

userlocalhost (approved):
Thank you for your work 👍

PR #672, opened 2022-11-03
Implement import/export feature for role

syucream:
SSIA

image


hinashi (approved):
LGTM!

("role with this name already exists" error occurred when I imported the exported data.)

userlocalhost (approved):
Thank you for your important work !!

PR #675, opened 2022-11-09
Fixed failed mypy check

hinashi:
CI became an error due to the version upgrade of mypy.


userlocalhost (approved):
Thank you for fixing this problem 👍

PR #676, opened 2022-11-09
Fixed the problem that ES registration process takes time

hinashi:
Changed method because search_entries makes elasticsearch high load.
get_es_document speeded up by using select_related.


userlocalhost (approved):
Thank you very much for this great performance improvement of register_es()

PR #682, opened 2022-11-14
Changed not to output if referral is not specified in search api

hinashi:
In the previous release, EntrySearchAPI was changed to output referral.
Depending on the Entry, the number of referrals will be large.
Reverted not to output if there is no referral parameter.


userlocalhost (approved):
Thank you for fixing my degradation implementation... lol

PR #686, opened 2022-11-17
Fixed elasticsearch nested object count limit error

hinashi:
After upgrading Elasticsearch, an error occurred in registering some entries.
```
RequestError(400, 'mapper_parsing_exception', {'error': {'root_cause': [{'type': 'mapper_parsing_exception', 'reason': 'The number of nested documents has exceeded the allowed limit of [10000]. This limit can be set by changing the [index.mapping.nested_objects.limit] index level setting.'}], 'type': 'mapper_parsing_exception', 'reason': 'The number of nested documents has exceeded the allowed limit of [10000]. This limit can be set by changing the [index.mapping.nested_objects.limit] index level setting.'}, 'status': 400})
```

It looks like Elasticsearch7 added a limit on the number of nested objects.
https://www.elastic.co/guide/en/elasticsearch/reference/7.0/breaking-changes-7.0.html#limit-number-nested-json-objects

### DIFF
--- a/airone/lib/elasticsearch.py
+++ b/airone/lib/elasticsearch.py
@@ -58,103 +58,110 @@ class ESS(Elasticsearch):
         self.indices.create(
             index=self._index,
             ignore=400,
-            body={
-                "mappings": {
-                    "properties": {
-                        "name": {
-                            "type": "text",
-                            "index": "true",
-                            "analyzer": "keyword",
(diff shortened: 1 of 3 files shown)


userlocalhost (approved):
LGTM 👍 Thank you for doing this important wok!!

PR #688, opened 2022-11-20
Avoid triggering to update state frequently

syucream:
Edit entity page is too slow because updating a state is  frequently triggered. I set `deps` to the states not to trigger so heavy.


hinashi (approved):
LGTM!

userlocalhost (approved):
It's awesome! Thank you for your great job!!

PR #693, opened 2022-11-27
Implement entry history page

syucream:
Renew design of the entry history page. Also support restoring an attribute value.

image


entry/api_v2/serializers.py
```diff
@@ -90,6 +90,61 @@ class EntryAttributeType(TypedDict):
     schema: EntityAttributeType
 
 
+class EntityAttributeTypeSerializer(serializers.Serializer):
```
syucream, line 93: It defines a similar serializers to subclasses of TypedDict defined before, because common TypedDict types will generate code has a curious class name if multiple serializers refer it.

syucream: It will break React-based custom views if those use these structures ...

userlocalhost (approved):
Thank you very much for your meaningful work!! 👍

hinashi (approved):
LGTM!

PR #695, opened 2022-11-30
Fixed export data in advenced search result

hinashi:
- Fixed not outputting in YAML export when there is no value in str, text type.
- Fixed an issue where boolean type was not output correctly in YAML export.
- Fixed an issue where role, array_role type was not output correctly in YAML export.
- Changed str, text type to not date format.

### DIFF
--- a/dashboard/tasks.py
+++ b/dashboard/tasks.py
@@ -51,7 +51,7 @@ def _csv_export(job, values, recv_data, has_referral):
             if (value is not None) and ("value" in value):
                 vval = value["value"]
 
-            if not value or "value" not in value or not value["value"]:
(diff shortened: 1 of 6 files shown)


dashboard/tasks.py
```diff
@@ -51,7 +51,7 @@ def _csv_export(job, values, recv_data, has_referral):
             if (value is not None) and ("value" in value):
                 vval = value["value"]
 
-            if not value or "value" not in value or not value["value"]:
+            if not value or "value" not in value or value["value"] is None:
```
userlocalhost, line 54: Thank you for this meaning full change 👍

userlocalhost (approved):
It's a great work!! Thank you for your meaningful work!!

PR #697, opened 2022-12-02
Prepare for added django-simple-history library

hinashi:
django-simple-history
https://django-simple-history.readthedocs.io/en/latest/

```
from pprint import pprint
from entity.models import Entity

entity = Entity.objects.first()
entity.history.count()

for history in entity.history.all():
  pprint(vars(history))

for history in entity.history.all():
  history.attrs.count()
```

### Show diffing
This is a simple syntax of showing difference between last HistoryRecord and the before last one.
```
def show_diffing(instance, offset=0):
  (before_last, last) = list(reversed(instance.history.order_by("history_id")))[offset:offset+2]
  for change in before_last.diff_against(last).changes:
      print("{} changed from {} to {}".format(change.field, change.old, change.new))
```

How to use this
```
```


userlocalhost (approved):
LGTM 👍

syucream (approved):
LGTM

PR #698, opened 2022-12-03
Fix null values on the entry export format

syucream:
Fix a value refer to deleted referral entries which will be null on the entry export feature, on the rule:

- entry
  - just have a null value
- array entry
  - remove elements have null value
- named entry
  - remove elements have empty name **and null value**
- array named entry
  - remove elements have empty name **and** null value


hinashi (commented):
It seems that the entry is not displayed if there is no name in named_entry, array_named_entry.
image

```
- entity: 削除確認
  entries:
  - attrs:
    - name: entry
      value: '20220312'
    - name: array_entry
      value:
      - '20220312'
    - name: named_entry
      value: {}
    - name: array_named_entry
      value: []
    name: test
```

syucream:
@hinashi It should be fixed!

Note: exported yaml file will have `? `, a complex mapping key written by pyyaml, e.g.
```
    - name: named_entry
      value:
        ? ''
        : bbb
```

https://github.com/yaml/pyyaml/issues/401

hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for your work !!

PR #700, opened 2022-12-05
Added link for switch design

hinashi:
Added a link in the header so you can switch designs.
image

image


userlocalhost (approved):
It's a good wire to switch new world!

PR #703, opened 2022-12-07
remove unused components and API client code

syucream:
- import entity/entry page component has been replaced with modals
- some API client V1 code has been replaced with V2's


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for your refactoring 👍

PR #709, opened 2022-12-15
Fixed edit entry page of customview

hinashi:
An error occurs when adding an attribute in a custom view.


userlocalhost (approved):
Thank you for your prompt work!!

PR #710, opened 2022-12-15
Fixed path of referral entry

hinashi:
Changed the URL of the reference entry.


frontend/src/components/entry/EntryReferral.tsx
```diff
@@ -102,7 +102,7 @@ export const EntryReferral: FC<Props> = ({ entityId, entryId }) => {
           >
             <ListItemButton
               component={Link}
-              to={entryDetailsPath(entityId, entry.id)}
+              to={entryDetailsPath(entry.schema.id, entry.id)}
```
userlocalhost, line 105: 👍

userlocalhost (approved):
Thank you very much for your important work!! 👍

PR #716, opened 2022-12-20
Enable to simple-search on AppBar

syucream:
Enable to perform simple search from the AppBar to allow users to quick access to some resources.

image


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for your prompt work!

PR #718, opened 2022-12-21
Fixed a bug of Entry.search_entries() when None is passed at the user parameter

userlocalhost:
It raise an exception when None is passed to the 1st argument (user). It expects to return result that is equivalent with the case when user who is deserved with readable permission.


hinashi (approved):
LGTM!

PR #723, opened 2022-12-23
Support pagination on advanced search result

syucream:
Support pagination on advanced search result page to dig a large amount of search results.

image

### DIFF
--- a/airone/lib/elasticsearch.py
+++ b/airone/lib/elasticsearch.py
@@ -809,6 +809,7 @@ def make_search_results(
     hint_attrs: List[Dict[str, str]],
     hint_referral: Optional[str],
     limit: int,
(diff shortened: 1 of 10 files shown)


airone/lib/elasticsearch.py
```diff
@@ -809,6 +809,7 @@ def make_search_results(
     hint_attrs: List[Dict[str, str]],
     hint_referral: Optional[str],
     limit: int,
+    offset: int = 0,
```
syucream, line 812: Ideally its better to paginate that before making search results, but it just supports limit-offset pagination here as the first step.

hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for an important work!!

PR #724, opened 2022-12-26
Autofocus at simple search box on dashboard page

syucream:
The dashboard page mainly provides simple-search feature, so `autofocus` at the text field makes usability better.
ref. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you so much for your good work 👍

PR #725, opened 2022-12-26
Make attribute name field shorter

syucream:
The entry attribute name field looks too long. I believe its better to see.

- before

image

- after

image


hinashi (approved):
LGTM!

userlocalhost (approved):
It's a good work for improving UX.

PR #728, opened 2022-12-27
Show entity name on simple search result

syucream:
It provides information which entity each entry is belonging to on search result page.
image


hinashi (approved):
LGTM!

userlocalhost (approved):
It's a nice change 👍 Thank you for a good work!

PR #731, opened 2023-01-04
Fixed header size

hinashi:
before:
image

after:
image


userlocalhost (approved):
Thank you for blushing up design 👍

PR #732, opened 2023-01-04
Fixed exact match search results were excluded in simple search

hinashi:
In search on TOP page, the entries are sorted by entry name.
Since the target entry may be excluded from the results,
Change to prioritize entries that exactly match entry name.


userlocalhost (approved):
It's a good change to improve UX 👍

PR #737, opened 2023-01-10
Fixed error in advanced search for all entities

hinashi:
If specify multiple entities in advanced search,
Limit parameters can be negative.
`Negative indexing is not supported`

### DIFF
--- a/airone/lib/elasticsearch.py
+++ b/airone/lib/elasticsearch.py
@@ -858,7 +858,7 @@ def make_search_results(
     hit_entries = Entry.objects.filter(id__in=hit_entry_ids, is_active=True)
 
     hit_infos: Dict = {}
-    for entry in hit_entries[offset : offset + limit]:
+    for entry in hit_entries[offset : max(0, offset + limit)]:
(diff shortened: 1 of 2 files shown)


airone/lib/elasticsearch.py
```diff
@@ -858,7 +858,7 @@ def make_search_results(
     hit_entries = Entry.objects.filter(id__in=hit_entry_ids, is_active=True)
 
     hit_infos: Dict = {}
-    for entry in hit_entries[offset : offset + limit]:
+    for entry in hit_entries[offset : max(0, offset + limit)]:
```
userlocalhost, line 861: It's a very smart fixing 👍

userlocalhost (approved):
I could catch up with why `limit` could be negative value. That is because [this processing](https://github.com/dmm-com/airone/blob/5dbf9997967111e333fdb07e316e5611cee9d6f5/entry/models.py#L2131) that prevent to adding result exceeds limit, and it must be necessary processing. So this fixing is excellent 👍

PR #738, opened 2023-01-14
Some improvements on user edit page

syucream:
- Disable reset buttons if user is not created
- Stop to crash on password reset modal if current password is invalid
- Validate input values on user edit form

image


hinashi (approved):
LGTM!

userlocalhost (approved):
Its'a good work! Thank you for blushing up UI 👍

PR #739, opened 2023-01-14
Show empty string for jobs doesn't have target

syucream:
To avoid failing to resolve target name, like:

image


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for fixing a bug 👍

PR #743, opened 2023-01-17
Fixed wrong ACL criterion for copy and delete processing

userlocalhost:
In the previous implementation, user who has writable permission can't copy and delete Entry. This commit fixed to allow it.


hinashi (approved):
LGTM!

PR #744, opened 2023-01-19
Fixed failed build

hinashi:
This is because the build failed in the local environment.
Due to a react version upgrade.


userlocalhost (approved):
Thank you for fixing 👍

PR #745, opened 2023-01-20
Reverted entity history apiv2

hinashi:
Reverted EntityHistoryAPI due to incompleteness.


userlocalhost (approved):
Thank you for noticing this important thing 👍

PR #747, opened 2023-01-23
Added processing to be able to configure sort-order at list Entry page.

userlocalhost:
User can change sort-order of Entry-list page by this PR.


hinashi (approved):
LGTM!

PR #750, opened 2023-01-26
The first step of creating service layer

userlocalhost:
This commit creates service layer. For starters, this separate service processing of Edit-Entry from pages/EditEntryPage.tsx.


hinashi (approved):
LGTM!

PR #751, opened 2023-01-27
Fixed failed uitest for v3.38.0

hinashi:
Fixed:
Frontend build failure problem
Entry deletion failure problem


userlocalhost (approved):
Thank you for covering my loss 🙏

PR #755, opened 2023-01-29
Improve role form to simplify, safe and human friendly more, with RHF and validation module

syucream:
I'd like to improve current form implementations to make easy to handle client side validation errors/server side errors, optimize performance, etc.
As the first step, I introduce RHF(React Hook Form) and a validation module to perform that on rewriting role form.

UPDATES
- Make it faster with RHF because it reduces the number of re-render!
- Remove some editing status states. Alternatively use some states provided by RHF `useForm()` like `isSubmitSuccessful`
- Notify server side errors via snackbar or each form fields.
  - Currently, notify `non_field_errors` via snackbar, field errors via each field, like this:

image

image

TODO
- [x] perform client side validation more effectively
  - It's better to consider using a validation library like zod, but describing `Controller.rules` once for now
- [x] handle server side errors


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for installing state-of-art technology 👍

PR #758, opened 2023-02-02
Changed to use DjangoContext instead of login page static descriptions

userlocalhost:
This fixed a problem that some component won't be respond at the login page.


hinashi (approved):
LGTM!

PR #759, opened 2023-02-02
Fixed anchor link in edit entry

hinashi:
Fixed an issue where clicking an anchor link after editing an attribute would show a prompt in edit entity page.


userlocalhost (approved):
It's great 👍 Thank you for your prompt work and complementing #761.

PR #761, opened 2023-02-02
Fixed problem to prevent showing wrong Prompt message

userlocalhost:
The problem would be happend when React page re-rendering processing run before applying changing state variable "submitted". This PR fixed it to guarantees moving page after changing state "submitted".


hinashi (approved):
LGTM!

PR #762, opened 2023-02-03
Fixed an error when there was no object type value in the entry details

hinashi:
In the current implementation, asObject, asGroup, and asRole may become undefined.


userlocalhost (approved):
LGTM 👍 Thank you very much for your prompt workaround work!!

PR #769, opened 2023-02-11
Filter only ancestors and others not to break group tree

syucream:
Filter ancestors and others as selectable groups for parent group on group form because specifying such groups as a parent breaks group tree structure.

In this PR, if we have such a group tree:

image

`group100` can select only `group1` as a parent (`group100`, itself and `group1000`, a descendant are not selectable).

image


hinashi (approved):
LGTM!

userlocalhost (approved):
Good catch!! Thank you for improving feature, and also separating service layer processing.

PR #772, opened 2023-02-13
Added customview in delete entry api

hinashi:
Added to the delete entry API because custom views may want to do some processing before deletion.


userlocalhost (approved):
Thank you very much for enhancing feature 👍

PR #773, opened 2023-02-14
Fixed permissions for entry operations

hinashi:
- Entity
  - Entry create: Writable
- Entry
  - get: Readable
  - update: Writable
  - delete: Writable
  - restore: Writable
  - copy: Writable


userlocalhost (approved):
Thank you for fixing implementation accompanied with changing ACL policy 👍

PR #774, opened 2023-02-15
Refactored django test case

hinashi:
Aserrt statement completion does not work when override_settings is used in decorators.


userlocalhost (approved):
Thank you for refactoring 👍

PR #776, opened 2023-02-16
Fixed pagination error in entry history apiv2

hinashi:
Fixed the following exception errors.
```
raised exception:
'LimitOffsetPagination' object has no attribute 'request'

  File "/airone/entry/api_v2/views.py", line 504, in list
    return self.paginator.get_paginated_response(serializer.data)
```


userlocalhost (approved):
It's a good hard work 👍

PR #778, opened 2023-02-21
Fixed entry history API for openapi generator

hinashi:
The build fails when you run generate:client.
Changed the function name because OpenAPI generator does not handle pagination unless the View function name is list.


userlocalhost (approved):

PR #782, opened 2023-02-24
Fixed search chain API when only one filtering hit

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for maintaining it 👍

PR #768, opened 2023-02-11
Try using Zod to perform client side validation

syucream:
Validate role form state on client side, and feedback errors, like:

image

PR #784, opened 2023-02-24
Fix UserForm with RHF and Zod

syucream:
Improve `UserForm` to make it more lightweight, user-friendly with RHF(React Hook Form) and Zod, like `RoleForm`.
Its similar modification to https://github.com/dmm-com/airone/pull/755 and https://github.com/dmm-com/airone/pull/768 .

As the result we got a chance to report client/server side validation errors:

image

image

And also I introduce to control to show/hide form fields based on login user (if its a superuser, if its user-itself).


hinashi (approved):
LGTM!

userlocalhost (approved):
It must be good feature for both user and software engineer 👍

PR #785, opened 2023-02-27
Changed to increase the maximum value of SearchChainAPI

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for good turning

PR #792, opened 2023-03-14
Added django-environ library

hinashi:
Enabled to change settings with environment variables.


userlocalhost (approved):
It's a great work to make AirOne more systematic!! Thank you very much for your hard work!!

PR #794, opened 2023-03-15
Fixed a bug to be able to edit Entry, Attribute and EntityAttr's ACL without Full permission

userlocalhost:
Previously there is a bug that following AirOne instance could be editable by user how doesn't have Full control permission for an Entity (e.g. Readable permission)
* Entry
* Attribute
* EntityAttr

That is because `get_obj_with_check_perm()` doesn't care about Entity's permission, but only checks specified ACLBase object's permission.

This PR fixed it by caring about hierarchical structure's permission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### Changed
 
 ### Fixed
+* Fixed a bug to be able to edit Entry, Attribute and EntityAttr's ACL without Full permission
+  Contributed by @userlocalhost
 
 ## v3.43.0
 
--- a/airone/lib/http.py
+++ b/airone/lib/http.py
@@ -48,41 +48,14 @@ def get_obj_with_check_perm(user, model, object_id, permission_level):
         return (None, HttpResponse("Failed to get entity of specified id", status=400))
 
     # only requests that have correct permission are executed
-    if not user.has_permission(target_obj, permission_level):
+    airone_instance = target_obj.get_subclass_object()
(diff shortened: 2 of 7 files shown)


hinashi (commented):
Since has_permission is also the same process, I thought it would be better to fix that one.
https://github.com/dmm-com/airone/blob/f72313864d074e3144fb54e1e1b93d2d674f5904/user/models.py#L88-L93

userlocalhost:
> Since has_permission is also the same process, I thought it would be better to fix that one.
> 
> https://github.com/dmm-com/airone/blob/f72313864d074e3144fb54e1e1b93d2d674f5904/user/models.py#L88-L93

Thank you very much for giving a great review. You are right!! It's reasonable to aggregate ACL checking processing to User.has_permission() because it has already had its processing as you said 👍 
I fixed it [this commit](https://github.com/dmm-com/airone/pull/794/commits/77b3d3983d8eda49eff400f9f88e9c77323fd212) !

userlocalhost:
## Remaining work
* Attribute ACL is editable when Entry has writable permission (ACL couldn't be editable)
(and so on)

### NOTE
Hierarchical ACL policy
* Attribute < Entry, EntityAttr, Entity
* Entry < Entity
* EntityAttr < Entity

userlocalhost:
I've fixed following problem.

> Attribute ACL is editable when Entry has writable permission (ACL couldn't be editable)

hinashi (approved):
LGTM!

PR #795, opened 2023-03-16
Added breadcrumbs component for entity and entry

hinashi:
I made the frequently used breadcrumbs a component.


userlocalhost (approved):
Thank you very much for your hard work 👍

PR #802, opened 2023-04-04
Fixed connection error when rabbitmq is scaled

hinashi:
https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-broker_heartbeat

If there is no heartbeat, reconnection will not be performed and a connection error will occur.


userlocalhost (approved):
Thank you very much for setting important parameter 👍

PR #805, opened 2023-04-09
Refactoring ReferralLikeAutocomplete

syucream:
Current `ReferralLikeAutocomplete` used by object-like, group, role attribute value field. In these use cases, group and role simply select fixed options, but object-like has more complex logic, something like incremental search. So I

- separate group/role and object-like usages
- improve Autocomplete object-like specific usage (see below video)

https://user-images.githubusercontent.com/191684/230753415-da03e70a-3705-4337-a99b-e2b7bef27c72.mov


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for improving in accordance with current data context.

PR #809, opened 2023-04-15
Separate not-named and named object fields

syucream:
The current `ObjectAttributeValueField` has common component for object, array-object, named-object, array-named-object types. I think its "too common" component that will be hard to improve in the future. So I separate it on use cases (non-named types and named types).


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for refactoring to be more maintainable 👍

PR #812, opened 2023-04-18
Added django storage for use cloud storage

hinashi:
Job export files can now be saved to Amazon S3.
This is need to use an NFS server in clustering configurations.


userlocalhost (approved):
It's good work to be cloud native configuration 👍

PR #814, opened 2023-04-21
Updated package-lock.json

hinashi:
Updated because the npm ci command is error.


userlocalhost (approved):
Thank you for fixing it 👍

PR #815, opened 2023-04-21
Fixed edit entries when special characters are used in attribute names

hinashi:
There is a problem that the value cannot be updated with ReactHookForm if the EntityAttr name contains "|".
https://github.com/react-hook-form/react-hook-form/issues/5441

Changed key in ReactHookForm to id of EntityAttr.


syucream (approved):

userlocalhost (approved):
Thank you for fixing 👍

PR #816, opened 2023-04-23
Fail early on generate spec.yaml

syucream:
`spectacular` command skips some spec generation if errors. Usually we, developers expect the generation command succeeds we get files.
To realize the failure of file generation, it fails early. For e.g.

```
 $ npm run generate:custom_client

> airone@1.0.0 generate:custom_client
> export DJANGO_CONFIGURATION=DRFSpectacularCustomView && ./tools/generate_client.sh

Error(s) occur on generating OpenAPI spec from API implementation:
[ERROR] asctime:2023-04-23 14:42:26,382 module:urls ....
 $
```


userlocalhost (approved):
Thank you very much for making clever solution to keep backward compatibility and increase maintainability 👍

hinashi (approved):
LGTM!

PR #818, opened 2023-04-24
Fixed a bug of /api_v1/entry/search_chain API implementation

userlocalhost:
This unables to filter intermediate search results with Entry name at the "refers" condition.


hinashi (approved):
LGTM!

PR #819, opened 2023-04-24
Fixed the order of attributes in the entry edit to index order

hinashi:
Changed the order of attributes to index order in Edit Entry.


userlocalhost (approved):
Thank you for fixing its problem 👍

PR #822, opened 2023-05-01
Changed to disable botocore because datadog trace failed

hinashi:
Disabled botocore tracing as it gives an error.
botocore is used in connecting to AWS s3.


syucream (approved):

userlocalhost (approved):
Thank you for updating it 👍

PR #824, opened 2023-05-07
Hide entry attribute if user isn't allowed to show it

syucream:
A similar change to https://github.com/dmm-com/airone/pull/794, for only entry page. Check `isReadable` that indicates user can read an attribute to determine if showing the value on the frontend.

image


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for implementing this feature 👍

PR #825, opened 2023-05-09
Fixed the process of excluding attributes in entry editing

hinashi:
Added Exclude Target Attributes from Entity Info.


userlocalhost (approved):
Thank you for making important work 👍

PR #829, opened 2023-05-12
Fixed not being able to add array_string in entry edit

hinashi:
Changed to an object because useFieldArray cannot add "" as a value.

```
AsIs: ArrayString: ["aaa", "bbb"]
ToBe: ArrayString: [{value: "aaa"}, {value: "bbb"}]
```


syucream (approved):

userlocalhost (approved):
Thank you for fixing problem 👍

PR #831, opened 2023-05-12
Fixed error when date attribute value is None in advanced search

hinashi:
AsString is of type string and was not expecting to return None, which caused an error.


syucream (approved):

userlocalhost (approved):
Thank you for fixing that principle problem 👍

PR #833, opened 2023-05-13
Hide disallowed value explicitly on advanced search result

syucream:
Hide disallowed value by entity attr explicitly on advanced search result page.
image

### DIFF
--- a/airone/lib/elasticsearch.py
+++ b/airone/lib/elasticsearch.py
@@ -891,6 +891,9 @@ def make_search_results(
             else:
                 ret_attrinfo = ret_info["attrs"][attrinfo["name"]] = {}
 
+            ret_attrinfo["is_readable"] = True
+            ret_attrinfo["type"] = attrinfo["type"]
+
             # if target attribute is array type, then values would be stored in array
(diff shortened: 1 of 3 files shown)


airone/lib/elasticsearch.py
```diff
@@ -891,6 +891,9 @@ def make_search_results(
             else:
                 ret_attrinfo = ret_info["attrs"][attrinfo["name"]] = {}
 
+            ret_attrinfo["is_readable"] = True
+            ret_attrinfo["type"] = attrinfo["type"]
```
syucream, line 895: If missing to assign type value (for e.g. `ret_attrinfo["is_readable"] = False; continue` like in this file), other part expects it has type field fails, like
```
...
  File "/Users/okubo-ryo/repos/dmm-com/airone/entry/api_v2/views.py", line 335, in post
    "type": attr["type"],
KeyError: 'type'
```

userlocalhost (approved):
Thank you for fixing that problem 👍

hinashi (approved):
LGTM!

PR #841, opened 2023-05-25
Fixed deleting entries in entry edit

hinashi:
image


userlocalhost (approved):
Thank you very much for solving problem.

PR #843, opened 2023-05-26
Updated datadog apm settings

hinashi:
Send data to host IP instead of localhost in k8s environment.
Deleted the setting because it can be set with the environment variable DD_AGENT_HOST.


userlocalhost (approved):
Thank you for good change for modernization 👍

PR #844, opened 2023-05-26
Fixed deleting entries in entry edit

hinashi:
There was an error in the following PR and it has been fixed.
https://github.com/dmm-com/airone/pull/841


userlocalhost (approved):
Thank you for fixing it 👍

PR #846, opened 2023-05-28
Improve AdvancedSearchAPI with DRF

syucream:
Improve AdvancedSearchAPI with DRF serializers and use auto-generated, typed client on the frontend side.


entry/api_v2/views.py
```diff
@@ -340,12 +309,21 @@ def _get_typed_value(type: int) -> str:
...
-                    entry["attrs"][name]["value"]["asString"] = ""
+                    entry["attrs"][name]["value"]["as_string"] = ""
+
+        serializer = AdvancedSearchResultSerializer(
+            data={"count": resp["ret_count"], "values": resp["ret_values"]}
+        )
+
+        # TODO validate response data strictly, like below.
+        # it'll fail because the data format will be different with EntryAttributeValueSerializer
+        # serializer.is_valid(raise_exception=True)
```
syucream, line 323: It'll cause validation errors in some case. For e.g. `... "value": {"asRole": {"id": "", "name": ""}}, ...` isn't acceptable because its illegal in entry attribute value serializer. We need to define advanced search specific serializers, or fix types of values to suit the serializer.

For now it puts on ice, this PR improves type safely with typed client code.

userlocalhost (approved):
Thank you for your good work 👍

PR #851, opened 2023-06-05
Fixed an error when POST parameter is incorrect format

hinashi:
There was a problem that an exception error occurred if the POST parameter was anything other than dictionary format.


userlocalhost (approved):
Good catch 👍 Thank you very much for fixing!!

PR #853, opened 2023-06-10
Install ruff as an alternative of flake8, autoflake and isort

syucream:
There're many devtools around Python. And actually we use `flake8`, `autoflake`, `isort`, ... It helps but sometimes confuses us.

Recently a new tool `Ruff` looks great as a replacement for these tools. Its so fast, supports auto-fix feature.
https://gihyo.jp/article/2023/03/monthly-python-2303

For now I'd like to just introduce it, doesn't replace existing tools. It it works well, I will remove flake8, autoflake and isort.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,11 +66,11 @@ jobs:
             chmod +x codecov
             ./codecov -t ${CODECOV_TOKEN}
 
-  pep8:
+  staticchecks:
     docker:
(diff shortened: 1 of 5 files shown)


hinashi (approved):
LGTM!

userlocalhost (approved):
It's a great work!! Thank you for refactoring service!!

PR #856, opened 2023-06-15
Fixed not showing attributes without permission

hinashi:
SSIA


syucream (approved):

userlocalhost (approved):
Thank you for fixing its bug 👍 It's a good work!

PR #857, opened 2023-06-16
Fixed duplicate attributes in complement_attrs

hinashi:
Added database constraints to prevent duplicate attributes.
Requires execution of the following SQL

```
ALTER TABLE `entry_attribute` ADD CONSTRAINT `unique_attribute` UNIQUE (`parent_entry_id`, `schema_id`);
```

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -447,6 +447,11 @@ class Attribute(ACLBase):
     schema = models.ForeignKey(EntityAttr, on_delete=models.DO_NOTHING)
     parent_entry = models.ForeignKey("Entry", on_delete=models.DO_NOTHING)
 
+    class Meta:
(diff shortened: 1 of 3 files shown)


entry/models.py
```diff
@@ -447,6 +447,11 @@ class Attribute(ACLBase):
     schema = models.ForeignKey(EntityAttr, on_delete=models.DO_NOTHING)
     parent_entry = models.ForeignKey("Entry", on_delete=models.DO_NOTHING)
 
+    class Meta:
```
userlocalhost, line 450: How elegant this fix is 👍

userlocalhost (approved):
Thank you very much for fixing that traditional problem with a elegant solution 👍

PR #858, opened 2023-06-19
Fixed exporting nothing when specifying all entities

hinashi:
- Added is_all_entities parameter to advanced search export.
- Fixed pagination in advanced search page


userlocalhost (approved):
Thank you for your prompt fixing that problem 👍

PR #859, opened 2023-06-20
Fixed high load on daily jobs when there are many entities

hinashi:
Changed to perform Elasticsearch registration process in the foreground.


userlocalhost (approved):
LGTM 👍

PR #864, opened 2023-06-30
Fixed pagination problem in advanced search

hinashi:
There was a bug in the limit and offset parameters.


syucream (approved):

userlocalhost (approved):
Thank you for your prompt work 👍

PR #867, opened 2023-07-02
Disallow blank string as entry name

### LINKED ISSUE #627: It is possible to create an entry with a blank name
# Problem

The name of the entry must be entered on the front end, but an empty entry can be created by including only spaces (half-width` ` or double-byte`　`) in the name.
Entries of including only spaces are same name, so `    ` and `　　　　　 ` are duplicate.

- The case of empty name(normal)
スクリーンショット 2022-09-15 13 01 02

- Include double-byte spaces entry (clickable)
スクリーンショット 2022-09-15 13 01 22

- Include only half-width spaces entry(non-clickable)
スクリーンショット 2022-09-15 13 02 16

- Empty name entry
スクリーンショット 2022-09-15 13 22 29

syucream:
fix https://github.com/dmm-com/airone/issues/627

Deny blank as entry name on both the current UI and the new UI.

image

image

### DIFF
--- a/entry/views.py
+++ b/entry/views.py
@@ -194,7 +194,11 @@ def create(request, entity_id):
 
 @http_post(
     [
-        {"name": "entry_name", "type": str, "checker": lambda x: x["entry_name"]},
+        {
+            "name": "entry_name",
+            "type": str,
+            "checker": lambda x: (x["entry_name"] and len(x["entry_name"].strip()) > 0),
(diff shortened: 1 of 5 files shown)


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for fixing a critical problem 👍

PR #873, opened 2023-07-07
Added react-infinite-scroller type package

hinashi:
Added because type checking fails.


syucream (approved):

userlocalhost (approved):
Thank you for fixing it 👍

PR #874, opened 2023-07-08
Support TEXT_NOT_CONTAINED filtering on advanced search result

syucream:
Enable to exclude some entries have attributes specified by keyword:
image


airone/lib/elasticsearch.py
```diff
@@ -40,6 +41,26 @@ class AdvancedSearchResults(TypedDict):
     ret_values: List[AdvancedSearchResultValue]
 
 
+class FilterKey(enum.Enum):
```
userlocalhost, line 44: Good to reduce frontend implementation by the "generate client" 👍

userlocalhost (approved):
It's a great expanding 👍 And how great you are!!

hinashi (approved):
LGTM!

PR #876, opened 2023-07-08
Replace ACLObjType with enum

syucream:
Replace some types for ACL with enum to make it reusable, type-safe.


hinashi (approved):
LGTM!

userlocalhost (approved):
It's a good fixing that makes implementation better for maintainability. Thank you very much 👍

PR #879, opened 2023-07-13
WIP: Added advanced search api from db

hinashi:


userlocalhost:
Following are needed @userlocalhost 
- [ ] Add tests
- [ ] Add script to set value to each Permission instances that have already been created from its codename field.

PR #866, opened 2023-07-02
Fix/update openapi generator

syucream:
Update openapi-generator-cli from v5.4.0 to v6.6.0.
https://github.com/OpenAPITools/openapi-generator/releases/tag/v6.6.0

The core change is here.
https://github.com/dmm-com/airone/compare/master...syucream:airone:fix/update-openapi-generator?expand=1#diff-ee761d3c9840dd8352d0677f0b31551e0e1d9085d4a9d010bdf6607f4d9dcf0f

PR #881, opened 2023-07-15
Follow changing error type on API client

syucream:
Since https://github.com/dmm-com/airone/pull/866 , the error handling doesn't work because the error type is changed from `Response` to `ResponseError`


hinashi (approved):
LGTM!

userlocalhost (approved):
Good catch! Thank you very much for fixing its problem 👍

PR #882, opened 2023-07-15
Support bulk delete entries

syucream:
Sometimes user want to delete multiple entries. For e.g. user mistakenly import entries multiple times then want delete them. So it newly support bulk delete specified entries on advanced search result:

https://github.com/dmm-com/airone/assets/191684/9ddb788c-79a3-430b-8114-6980db21e863


hinashi (approved):
LGTM!

userlocalhost (approved):
LGTM 👍

PR #883, opened 2023-07-15
Some improvements on advanced search

syucream:
- Fix variable name/type
- Add title of attr filter menu
- Have advanced search params formatter to centralize params generations
  - Currently we have some URL search params generation logic at some places, but its fragile and not so testable


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for blushing up this implementation 👍

PR #885, opened 2023-07-20
Fixed entry histoty page

hinashi:
- Changed the entry history API to return the previous attribute value ID.
- Fixed an error with some attribute values.
- Modified the design a little.


userlocalhost (approved):
Thank you very much for fixing potential problems 👍

PR #886, opened 2023-07-20
Fixed api error handling

hinashi:
Same problem as the PR below.
https://github.com/dmm-com/airone/pull/881


userlocalhost (approved):
Thank you very much for blushing up its implementation 👍

PR #891, opened 2023-07-22
Make extracting advanced search params more safe

syucream:
Formatting React values to URL query params is centralized at https://github.com/dmm-com/airone/pull/883
But extracting side is still independent, so the format/extraction will be broken on future changes.

So I centralize extraction logic on service layer, and restrict format/extraction to use predefined, shared key names (e.g. `is_all_entities`) to make it more safe.


frontend/src/components/entry/AdvancedSearchModal.tsx
```diff
@@ -12,7 +12,7 @@ import React, { Dispatch, FC, useState, SetStateAction } from "react";
 import { useHistory } from "react-router-dom";
 
 import { AdvancedSearchResultAttrInfoFilterKeyEnum } from "../../apiclient/autogenerated";
-import { formatAdvancedSearchPrams } from "../../services/entry/AdvancedSearch";
+import { formatAdvancedSearchParams } from "../../services/entry/AdvancedSearch";
```
syucream, line 15: fix a typo 😓

hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for your continuous improvements 👍

PR #892, opened 2023-07-22
Use predefined attribute value type

syucream:
The React UI has attribute value type itself. But ideally it should be shared with server side.
In this PR, it defines the types with `enum.Enum`, generate enum type values from the enum, finally replace duplicated types definitions with the generated.


frontend/src/services/DjangoContext.ts
```diff
@@ -29,23 +28,6 @@ export class DjangoContext {
     this.noteLink = context.note_link;
     this.version = context.version;
     this.user = context.user ? new User(context.user) : undefined;
-
```
userlocalhost, line 32: 👍

userlocalhost (approved):
Thank you very much for improving its implementation 👍

hinashi (approved):
LGTM!

PR #893, opened 2023-07-24
Fixed containers filling disk space in logs

hinashi:
In container environments such as k8s, logs should be output only to the console.


userlocalhost (approved):
Thank you for setting very important configuration 👍

PR #894, opened 2023-07-26
Fixed acl setting page

hinashi:
- Removed unused API response fields.
- Speed up ACL update process.
- Changed page after ACL update.
- Modified design a little.


userlocalhost (approved):
Thank you for fixing problems 👍

PR #896, opened 2023-07-29
Convert to Blob before calling APIV2 client method

syucream:
Probably, after https://github.com/dmm-com/airone/pull/866, import API calls should fails because the request body is redundantly stringify by the generated client code at https://github.com/dmm-com/airone/blob/4502a49d8440e8922227be56b581af4f1729f6a5/frontend/src/apiclient/autogenerated/runtime.ts#L206-L214.

So I bypass it, converts string body to Blob type.


userlocalhost (approved):
Thank you for adapting for new open-api version's behavior 👍

hinashi (approved):
LGTM!

PR #897, opened 2023-07-31
Fixed helm build error

hinashi:
MySQL's helm has a problem that the past version has not been released.

- mysql
https://mysql.github.io/mysql-operator/index.yaml
- elasticsearch
https://helm.elastic.co/index.yaml


userlocalhost (approved):
How great to solve this problem 👍

PR #898, opened 2023-08-03
Fixed acl history page

hinashi:
- Fixed a bug in ACL update API.
- Changed ACL role history to be more readable.
- Modified design a little.


userlocalhost (approved):
Thank you for your great work 👍

PR #900, opened 2023-08-03
Fixed problem that prohibited attribute to be edited is set to blank at Entry edit processing

### LINKED ISSUE #899: Prohibited attribute to access may be set to blank
An attribute's value would be set to blank when an user who doesn't have permission to write it edit an Entry that includes that Attribute.

[Reproducing procedure]
1. Create an Entity (E1) with Attribute (A).
2. Edit ACL of A to prohibit to be able to edit by only Role(R).
3. Create an Entry (e1) of E1 and set value to A by User(A) who belongs to R.
4. Update e1 by User(B) who **doesn't** belong to R.

Then, the value of e1.A will be blank.

userlocalhost:
Close: #899

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### Changed
 
 ### Fixed
+* Fixed problem that prohibited attribute to be edited is set to blank at Entry edit processing (#899).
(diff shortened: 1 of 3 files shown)


hinashi (approved):
LGTM!

PR #903, opened 2023-08-05
Enable to import objects have same name safely

syucream:
Currently the import entries by entry names. But the name will conflict with entries under the other entity.

For e.g., if we have an entry has object-like fields referring to `entry1`
```yaml
      - name: referral
        value: entry1
```
and the entity that the parent of the entry has both `entity1` and `entity` as the type of `referral`, and also both an entry under `entity1` named `entry1` and an entry under `entity1` named `entry1`. The name `entry1` in the yaml file is ambiguous, importing result of the yaml file is unstable.

To resolve the problem, I fix the yaml export format  a little to have an entity name explicitly. For e.g.
```yaml
      - name: referral
        value: 
          name: entry1
          entity: entity1
```
The yaml file specifies `entity1` as the parent of `entry1` so the importing result should be deterministic.

### DIFF
--- a/entry/api_v2/serializers.py
+++ b/entry/api_v2/serializers.py
@@ -694,7 +694,7 @@ class EntryImportEntitySerializer(serializers.Serializer):
         def _convert_value_name_to_id(attr_data, entity_attrs):
             def _object(
                 val: Optional[Union[str, dict]],
-                refs: list[ACLBase],
+                refs,
             ):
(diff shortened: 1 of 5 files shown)


userlocalhost (approved):
It's also great!! Thank you very much for its good work 👍

hinashi (approved):
LGTM!

PR #904, opened 2023-08-05
Fix entry copy and entry list pages

syucream:
- allow to resize copy form, and fix timing to leave the page after submitting to avoid noisy prompt
- show the number of entries per pages on entry list page

image


hinashi (approved):
LGTM!

userlocalhost (approved):
It's good works 👍 Thank you for fixing it !

PR #906, opened 2023-08-10
Fixed a bug not to be able to get AttributeValue instance

userlocalhost:
When an EntityAttr's name was changed after creating Entry that refers its EntityAttr.


syucream (approved):

hinashi (approved):
LGTM!

PR #912, opened 2023-08-14
Fixed advanced search page

hinashi:
- Modified the design a little.
- Fixed pagination issue when filtering.
- Fixed page transition when deleting an entry.


userlocalhost (approved):
Thank you for improving UI implementation 👍

PR #913, opened 2023-08-16
Fixed display when entry has no permissions

hinashi:
There was no logic for when an entry has no permissions in advanced search.


userlocalhost (approved):
It's a good work 👍 Thank you!!

PR #917, opened 2023-08-18
Fixed warning message in <a> tag

hinashi:
A warning is displayed if there is an a tag inside an a tag


syucream (approved):
oops, nice catch! 🙏

userlocalhost (approved):
Good work 👍

PR #920, opened 2023-08-19
Ignore E721 for now

syucream:
Its a bad line for ruff check but now necessary, so just ignoring for now.

PR #921, opened 2023-08-19
Use default destination

syucream:
We don't have `/index.html` and https://github.com/dmm-com/airone/deployments/github-pages navigates https://dmm-com.github.io/airone/. But currently we have pages under `docs` so accessible page URL is https://dmm-com.github.io/airone/docs/
The `docs/` directory is necessary? If not so, it looks better to remove the level. How about this?


syucream:
The error on staticchecks should be fixed by https://github.com/dmm-com/airone/pull/920

hinashi:
helm is published in index.yaml.
There is a problem that deploying hugo deletes the index.yaml if `/airone/` is the destination.
It may be possible to avoid it with options. exclude_assets? keep_files?
https://github.com/peaceiris/actions-gh-pages

Also, the following will need to be changed.
https://github.com/dmm-com/airone/blob/843e4f0c2f15d254b15ade3e4b051d2aa22a39e6/docs/config.yaml#L2

syucream:
fixed!

hinashi (approved):
LGTM!

.github/workflows/release-hugo.yml
```diff
@@ -24,4 +24,4 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./docs/public
-          destination_dir: docs
+          keep_files: true
```
userlocalhost, line 27: Thank you for solving following requirement with elegant solution 👍 
https://github.com/dmm-com/airone/pull/921#issuecomment-1687599150

userlocalhost (approved):
Thank you very much for its great work!!

PR #923, opened 2023-08-22
Upgrade drf_spectacular version

hinashi:
- Upgrade drf_spectacular version
- django.conf.urls.url is deprecated and changed.


userlocalhost (approved):
Thank you very much for upgrading it! 👍

PR #924, opened 2023-08-24
Changed value type for named_entry attribute

hinashi:
Changed the type of the named_entry attribute because there was a problem updating the array_named_entry_boolean attribute.


userlocalhost (approved):
Thank you for making processing to identify boolean type at ArrayNamedObject type 👍

PR #918, opened 2023-08-19
Finally use published API client package; goodbye repository-owned client!

syucream:
At https://github.com/dmm-com/airone/pull/902, we've finally got published npm package here.
https://github.com/dmm-com/airone/pkgs/npm/airone-apiclient-typescript-fetch/118338298

So, now we can replace code on `frontend/src/apiclient/autogenerated` which's simple, but painful mechanism with installing `@dmm-com/airone-apiclient-typescript-fetch`. I do:

- replace the existing import statements with the published package.
- describe how to use the new mechanism
- remove `frontend/src/apiclient/autogenerated`, goodbye! 👋 

As for https://github.com/dmm-com/airone/issues/834, I also describe how to work/use the mechanism.

PR #927, opened 2023-08-26
Enable to install dependencies with GITHUB_TOKEN

syucream:
It will resolve the npm-ci error on https://github.com/dmm-com/airone/pull/918.
I guess that the action runs workflows defined at `master` branch, not the feature's because of `pull_request_target` trigger. So at first we'll need to modify the workflow definition.


hinashi (approved):
LGTM!

> I guess that the action runs workflows defined at master branch, not the feature's because of pull_request_target trigger. So at first we'll need to modify the workflow definition.

Thank you for letting me know.

userlocalhost (approved):
LGTM 👍 Thank you for your good work 👍

PR #930, opened 2023-08-26
Show exact total number if the number exceeds 10000

syucream:
Current advanced search cannot show search result number 10000+ because of elasticsearch's default behavior.
https://www.elastic.co/guide/en/elasticsearch/reference/7.17/breaking-changes-7.0.html#track-total-hits-10000-default

But we want the exact total number, so it gives `track_total_hits` to get it.
For e.g.
image
image


hinashi (approved):
LGTM!

userlocalhost (approved):
How great it is !! 👍 Thank you very much!!

PR #936, opened 2023-08-31
Fixed failed install for mac

hinashi:
SSIA


userlocalhost (approved):
Thank you for tackling with this problem!! And it's great to find out a solution 👍

PR #939, opened 2023-09-01
Fixed failed generate_client

hinashi:
There was an issue where npm run generate:client would fail if a custom view was not set.
https://github.com/dmm-com/airone/blob/f3a575df9506d49418de1184135cce684edf2192/tools/generate_client.sh#L8


userlocalhost (approved):
Thank you for fixing it 👍

PR #942, opened 2023-09-03
Paginate on ES query, not the application side

syucream:
Currently advanced search can perform pagination, but the control is performed on the application side (mainly `airone/lib/elasticsearch.py`), not on elasticsearch query. So if we have many entries as advanced search result candidates, the performance is so heavy because airone fetches "all the results" as possible, then filter it.
To optimize it, I would switch to ES side pagination with appropriate `size` and `from` parameters. Actually I can get performance improvement as below:

### environment

- I have 10000+ entries
- ES runs on docker container with poor machine resource

### before

I set `MAXIMUM_RESULTS_NUM=1000` because the default (too large) value `500000` does not work well on my box.

image

### after

looks better than 'before''s

image

### DIFF
--- a/airone/lib/elasticsearch.py
+++ b/airone/lib/elasticsearch.py
@@ -877,12 +877,15 @@ def _make_an_attribute_filter(hint: AttrHint, keyword: str) -> Dict[str, Dict]:
     return {"nested": {"path": "attr", "query": {"bool": {"filter": cond_attr}}}}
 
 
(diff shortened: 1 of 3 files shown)


hinashi (approved):
LGTM!
It was very fast in my environment.

userlocalhost (approved):
How great it is!! This must reduce not only client wait time, but also workload of application worker 👍

PR #945, opened 2023-09-06
Fix bug of Entry, user, etc. menu to show deleted Entry just after deleting

sniper-fly:


hinashi (approved):
LGTM
Thanks all for the lighting fast response!

PR #948, opened 2023-09-09
Make advanced search result export v2 compatible with entry import v2

syucream:
The advanced search result export v1 outputs entry import v1 compatible data format, but v2 is not so. Its because the v2 export data format is same as v1's ...

### DIFF
--- a/airone/lib/http.py
+++ b/airone/lib/http.py
@@ -168,6 +168,7 @@ def render(request, template, context={}):
             "EXPORT_V2": JobOperation.EXPORT_ENTRY_V2.value,
             "RESTORE": JobOperation.RESTORE_ENTRY.value,
             "EXPORT_SEARCH_RESULT": JobOperation.EXPORT_SEARCH_RESULT.value,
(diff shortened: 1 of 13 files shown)


hinashi (commented):
The format of entry, named_entry, group, role, and date attributes was different from entry export.
Therefore the import will fail.

### advanced search result export
```
  - attrs:
    - name: entry
      value:
        entity: null
        name: ''
    - name: named_entry
      value:
        ? ''
        : entity: null
          name: ''
    - name: group
      value: ''
    - name: role
      value: ''
    - name: date
      value: '2023-01-01'
```
### entry export
```
  - attrs:
    - name: entry
      value: null
    - name: named_entry
      value: {}
    - name: group
      value: null
    - name: role
      value: null
    - name: date
      value: 2023-01-01
```

syucream:
Fixed how to handle no-value cases. How about that?

hinashi (approved):
LGTM! thanks for fixing.

userlocalhost (approved):

PR #949, opened 2023-09-09
Do not crash if refreshing jobs failed

syucream:
The recent jobs are automatically, periodically updated on background. Currently if the update failed, the new UI will crash. But I think the update process is not so critical, should keep to serve. (And I think it will occur frequently in development environment...)

After this modification, the update failures ignored:

image


hinashi (approved):
LGTM!

userlocalhost (approved):
LGTM 👍 Thank you very much for applying this problem with very reasonable solution.

PR #950, opened 2023-09-13
Fixed failed ci

hinashi:
SSIA


userlocalhost (approved):
Thank you for fixing it 👍

PR #951, opened 2023-09-14
Fixed an error when the limit value was large in entry search

hinashi:
SSIA


userlocalhost (approved):

PR #954, opened 2023-09-16
Check role editability before update/destroy in the new UI

syucream:
Currently deleting role is allowed by **non**-administrative users. Of course its bad. So I 1) fix the APIv2 to check user permissions correctly, and 2) do not show navigations to edit-like actions if login user doesn't appropriate permission to do that.

image


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.0.5",
+  "version": "0.0.6",
```
syucream, line 3: @userlocalhost @hinashi could you attach release label if its okay? 🙏

hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for your good work 👍

PR #959, opened 2023-09-22
Fixed exception error in entry import

hinashi:
An error occurred when an entry type had multiple reference entities.


syucream (approved):

userlocalhost (approved):
Thank you very for fixing with the smallest change 👍

PR #963, opened 2023-09-29
Add DataUtil.formatDate() that only returns date (without time) string value from Date instance

userlocalhost:
This commit distinguishes DataUtil.formatDateTime(), which is old formatDate() method, and DateUtil.formatDate() that only returns date (without time) value as string from Date instance.
It's useful for some custom views :)


hinashi (approved):
LGTM!

PR #964, opened 2023-09-29
Fixed required for mandatory attributes when renaming using entry update API

hinashi:
SSIA


syucream (approved):

userlocalhost (approved):
Thank you very much for doing this important work!! 👍

PR #968, opened 2023-09-30
Update Django to 4.2

syucream:
SSIA, to the next LTS version


.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ jobs:
         environment:
           discovery.type: single-node
           ES_JAVA_OPTS: -Xms512m -Xmx512m
-      - image: mysql:5.7.34
+      - image: mysql:8.1
```
syucream: Its the key point on updating Django to 4.2. It requires MySQL8+

hinashi (approved):
Sorry for the long wait 😿

PR #972, opened 2023-10-07
Refine props passed in advanced search params

syucream:
`SearchResultControlMenu` accepts filter conditions for **all** attributes but each component needs condition for specific attribute. I guess it occurs unnecessary re-rendering. So I refine props to get only necessary values.

### DIFF
--- a/frontend/src/components/entry/SearchResultControlMenu.tsx
+++ b/frontend/src/components/entry/SearchResultControlMenu.tsx
@@ -13,7 +13,7 @@ import {
 import { styled } from "@mui/material/styles";
 import React, { ChangeEvent, FC } from "react";
 
-import { AttrsFilter } from "../../services/entry/AdvancedSearch";
+import { AttrFilter } from "../../services/entry/AdvancedSearch";
 
 const StyledTextField = styled(TextField)({
   margin: "8px",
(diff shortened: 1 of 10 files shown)


frontend/src/components/entry/SearchResultControlMenu.tsx
```diff
@@ -24,75 +24,61 @@ const StyledBox = styled(Box)({
 });
 
 interface Props {
-  attrName: string;
-  attrsFilter: AttrsFilter;
+  attrFilter: AttrFilter;
```
syucream, line 27: Here is the key point; changed prop type from `Record<K, V>` to `V`.

hinashi:
https://github.com/dmm-com/airone/assets/5561786/41ab811a-8544-45d4-9037-ff744804ad96

However, it still took a long time to input.
This seems to occur when there are a large number of search results.
There are fewer results in the second half of the video, so it doesn't take much time to input.

The problem seems to be that the results are redrawn when the attrfilter state is updated.

syucream:
@hinashi I have additional improvement on https://github.com/dmm-com/airone/pull/972/commits/20876321698a9e9f5be0ff5513da48bcc956de29, could you try it?

hinashi (approved):
LGTM
Thank you! very fast response

userlocalhost (approved):
Thank you very much for great improvement!! It's a good work 👍

PR #974, opened 2023-10-07
Limit freqnent import jobs for same entity by default

syucream:
Some users accidentally import entries of the same entity, the import result sometimes will be unexpected state. To resolve it, I introduce frequent limitation for the entry import, and an import option to ignore the limitation by users explicitly.

image


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.0.6",
+  "version": "0.0.7",
```
syucream, line 3: @userlocalhost @hinashi could you attach the release label if its okay? 🙏

hinashi:
waiting for PR below to be merged.
https://github.com/dmm-com/airone/pull/975

userlocalhost (approved):
@syucream 
Thank you very much for your prompt work that we requested!! 

@hinashi
And thank you very much for working to pass CI processing.

PR #975, opened 2023-10-11
Specified python version

hinashi:
CI fails due to differences in python microversions.


userlocalhost (approved):
Good job! Thank you very much for fixing this difficult problem 👍

PR #978, opened 2023-10-15
Avoid unnecessary re-loading on entity selector

syucream:
To stop flicking the entity selector to improve UX.


hinashi (approved):
LGTM!

userlocalhost (approved):
It's an informative fixing. Thank you for teaching it 👍 
We understand that we should fix other implementation in this way.

PR #980, opened 2023-10-16
Fixed for header width would change depending on the presence or absence of a scroll bar

hinashi:
There was an issue where the header was shifted when opening the select box.
(e.g. Edit Entity page)
Fixed it by taking the width of the scrollbar into account.


userlocalhost (approved):
Thank you for fixing this difficult problem 👍 It's great!

PR #981, opened 2023-10-16
Fixed for downloaded files were text garbled

hinashi:
When using Excel on Windows, the preferred character code is SHIFT_JIS.
Since the desired character code differs depending on the user, we have made it possible to choose.


userlocalhost (approved):
How great change it is!! User must be satisfied with this feature 👍

PR #986, opened 2023-10-22
Enable to limit the number of records

syucream:
It enables to limit the number of records on:

- entity
- entity attributes per entity
- entry
- user
- group
- role

to prevent a user exhausts server resources.
The limitation is optional (`None`(default) means unlimited ) and we can specify it via settings with environment variables.

### DIFF
--- a/airone/settings_common.py
+++ b/airone/settings_common.py
@@ -2,7 +2,7 @@ import errno
 import logging
 import os
 import subprocess
-from typing import Any
+from typing import Any, Optional
 
 import environ
 from configurations import Configuration
@@ -432,3 +432,11 @@ class Common(Configuration):
                 "dd.span_id:%(dd.span_id)s",
             ]
         )
+
+    # Dynamic record number limitations on model level validation (None means unlimited)
+    MAX_ENTITIES: Optional[int] = env.int("AIRONE_MAX_ENTITIES", None)
+    MAX_ATTRIBUTES_PER_ENTITY: Optional[int] = env.int("AIRONE_MAX_ATTRIBUTES_PER_ENTITY", None)
+    MAX_ENTRIES: Optional[int] = env.int("AIRONE_MAX_ENTRIES", None)
+    MAX_USERS: Optional[int] = env.int("AIRONE_MAX_USERS", None)
+    MAX_GROUPS: Optional[int] = env.int("AIRONE_MAX_GROUPS", None)
+    MAX_ROLES: Optional[int] = env.int("AIRONE_MAX_ROLES", None)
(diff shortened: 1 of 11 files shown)


airone/settings_common.py
```diff
@@ -432,3 +432,11 @@ class Common(Configuration):
...
             ]
         )
+
+    # Dynamic record number limitations on model level validation (None means unlimited)
+    MAX_ENTITIES: Optional[int] = env.int("AIRONE_MAX_ENTITIES", None)
+    MAX_ATTRIBUTES_PER_ENTITY: Optional[int] = env.int("AIRONE_MAX_ATTRIBUTES_PER_ENTITY", None)
+    MAX_ENTRIES: Optional[int] = env.int("AIRONE_MAX_ENTRIES", None)
+    MAX_USERS: Optional[int] = env.int("AIRONE_MAX_USERS", None)
+    MAX_GROUPS: Optional[int] = env.int("AIRONE_MAX_GROUPS", None)
+    MAX_ROLES: Optional[int] = env.int("AIRONE_MAX_ROLES", None)
```
userlocalhost, line 442: That's great! These are things that we are wanting to implement 👍

userlocalhost (approved):
Thank you very much for your prompt work to implement what we want!

entity/models.py
```diff
@@ -73,6 +76,16 @@ def add_referral(self, referral):
...
 
+    def save(self, *args, **kwargs):
+        max_attributes_per_entity: Optional[int] = settings.MAX_ATTRIBUTES_PER_ENTITY
+        if (
+            max_attributes_per_entity
+            and EntityAttr.objects.filter(parent_entity=self.parent_entity).count()
+            >= max_attributes_per_entity
+        ):
+            raise RuntimeError("The number of attributes per entity is over the limit")
+        return super(EntityAttr, self).save(*args, **kwargs)
```
userlocalhost, line 87: It's a very smart solution. Thank you!!

userlocalhost:
@syucream 
Thank you very much for making this great feature with very simple implementation!

Could I add additional specification? That adds regulations to limit length of followings.

### Entry name
There is a bug that make you suprise when you save an Entry with specifying more than 200 characters as its name.

### AttributeValue

There is no limitation to store value at AttributeValue.value. By default, it can store big size data (up to 4GB) for each.

(c.f. https://docs.djangoproject.com/en/dev/ref/databases/#textfield-limitations)
(c.f. https://dev.mysql.com/doc/refman/8.0/en/string-type-syntax.html)

---
It's OK either adding it on this PR or making another PR.

hinashi (approved):
LGTM!

PR #987, opened 2023-10-22
Record change history on specific fields

syucream:
Currently the entity update always records history even if it has no changes. Its redundant, so I filter target fields for recording history.

### DIFF
--- a/entity/api_v2/serializers.py
+++ b/entity/api_v2/serializers.py
@@ -200,11 +200,11 @@ class EntitySerializer(serializers.ModelSerializer):
         if not is_created_entity:
             # record history for specific fields on update
             updated_fields: list[str] = []
-            if entity.name != validated_data["name"]:
-                entity.name = validated_data["name"]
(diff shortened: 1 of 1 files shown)


hinashi (approved):
LGTM!

entity/api_v2/serializers.py
```diff
@@ -192,10 +192,24 @@ def _update_or_create(
         is_toplevel_data = validated_data.pop("is_toplevel", None)
         attrs_data = validated_data.pop("attrs")
         webhooks_data = validated_data.pop("webhooks")
+
         entity: Entity
-        entity, is_created_entity = Entity.objects.update_or_create(
+        entity, is_created_entity = Entity.objects.get_or_create(
```
userlocalhost, line 197: (memo) I understand update_or_create() will update processing no matter how it is not updated. The variable `is_created_entity` will be True when a new Entity is created.

userlocalhost (approved):
Thank you for making good feature to save unnecessary history instances.

PR #988, opened 2023-10-22
Allow to submit if any fields are dirty

syucream:
Current the new UI allows to submit data that doesn't have any changes. Its unnecessary, and will cause redundant history record ( see also https://github.com/dmm-com/airone/pull/987 ).
So I enforce forms to have any changed fields on submit. Its easy to recognize it, just with `isDirty` from RHF.

https://github.com/dmm-com/airone/assets/191684/052fd8a5-cf59-47f8-b8d5-39543115fd21

### DIFF
--- a/frontend/src/pages/EditEntityPage.tsx
+++ b/frontend/src/pages/EditEntityPage.tsx
@@ -203,7 +203,7 @@ export const EditEntityPage: FC = () => {
       >
         <SubmitButton
           name="保存"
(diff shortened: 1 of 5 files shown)


userlocalhost (approved):
How great it is!! This feature saves our infrastructure!!
Thank you very much 👍

hinashi (approved):
LGTM!

PR #989, opened 2023-10-24
Fixed datadog's APM data could not be obtained

hinashi:
SSIA


userlocalhost (approved):

PR #990, opened 2023-10-26
Added drop down button in header

hinashi:


userlocalhost:
Added processing to be able to show user defined menu on the application navigation header like this.
スクリーンショット 2023-10-27 12 40 29

You can configure this menus from [here](https://github.com/dmm-com/airone/pull/990/files#diff-8caa08fb8da06b64e9daa43b0e31b3380fb8a5ee6810b3c8c34b66429fa1e545R220-R232)

PR #992, opened 2023-10-28
Validate some data length on FE side

syucream:
A follow-up update for https://github.com/dmm-com/airone/pull/986#issuecomment-1774337543

- Validate entry name on the new UI side.
- Validate max attribute value on the new UI side.
  - No our own limitation on DB side (so the default, 4GB length is applied), but there's a similar validation logic on APIv2
  - https://github.com/dmm-com/airone/blob/94953248d44171f34375de0eccf04cf697a96041/entry/api_v2/serializers.py#L222-L227
  - So for now, I append the same limitation on the FE side.

Any other preferred limitation?


hinashi (approved):
LGTM!

userlocalhost (approved):

PR #993, opened 2023-10-28
Enable to switch legacy UI

syucream:
The new ui will be standard UI for AirOne, and the legacy( current ) UI will be deprecated. Keeping to support the legacy UI can be costly for administrators. So I'd like to switch the path to legacy UI. It can be performed via `AIRONE_LEGACY_UI_DISABLED` environment variable on this PR.


hinashi (approved):
LGTM!

userlocalhost (approved):
How elegant to manage service generation 👍 Thank you very much for your great work!

PR #994, opened 2023-10-30
Updated MySQL helm repo

hinashi:
Fixed an error when releasing helm.
https://github.com/dmm-com/airone/actions/runs/6688401952


userlocalhost (approved):
LGTM 👍 Thank you for maintaining to catch up with MySQL official manifest.

PR #997, opened 2023-11-03
Add date/boolean type specific attr filter selector

syucream:
The attribute filter on the advanced search result page is not so user-friendly for now. Its hard to select date/boolean typed value on just a text field. So I introduce these type specific selectors if the types can be guesses appropriately.

image
image

As the limitation, type guess just uses the first element on search results. **The element sometimes doesn't have attribute type (for e.g. a date type attribute was appended after the first element created.)** To cover such a situation, we need to get more correct attribute types via API.


hinashi (approved):
LGTM
Easy and convenient to use!

userlocalhost (approved):
Thank you for your great work!! It must improve UX 👍

PR #999, opened 2023-11-03
Add a simple component test case with RTL

syucream:
SSIA


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for this important work that can test each components behavior using this control mock. It's a great first step 👍

PR #1003, opened 2023-11-09
Added trigger action model

hinashi:
## Summary
Added new feature to be able to update Entry automatically in association with user-defined condition (named `TriggerCondition` that describes which Attribute will be updated by specified value(`TriggerAction`) when specified Attribute(s) are updated with expected value(s)).


userlocalhost:
## Note

- [x] Use UpdateSerializer to update Entry and its related processings
- [x] Mkae processing to prevent updating loop of configured Entries

userlocalhost:
Up to this point, it's completed to implement feature of TriggerCondition and TriggerAction that changes configured Attribute values when an Entry is updated through APIv2.

PR #1005, opened 2023-11-12
i18n support experimentally

syucream:
i18n support with i18next
https://qiita.com/Syoitu/items/98343af1bd69e7c38283

image

The language is auto-detected based on browser language setting, like this in Chrome:
image

We can scale i18n setting by following steps:
1. Update i18n dictionary defined in `frontend/src/i18n/config.ts`
2. Replace each components to use the translation hook, like `frontend/src/components/Header.tsx`


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for making sophisticated processing to make i18n 👍

PR #1007, opened 2023-11-19
Enhance React FE tests

syucream:
- Add some test cases for entry form
- Remove enzyme because its not so maintained actively (especially React 17+ isn't supported officially)


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for making its great example for FE test. It makes us easy to develop rest of ones by your favor 👍

PR #1009, opened 2023-11-23
Reset ErrorBoundary on relocations

syucream:
SSIA, especially, to support escaping an error page by browser back navigation.

https://github.com/dmm-com/airone/assets/191684/7f7badd1-8e08-43d7-afbe-307c7a7c243d

### DIFF
--- a/frontend/src/AppRouter.tsx
+++ b/frontend/src/AppRouter.tsx
@@ -2,6 +2,7 @@ import React, { FC } from "react";
 import { RouteComponentProps } from "react-router";
 import { BrowserRouter as Router, Switch, Route } from "react-router-dom";
 
+import { ErrorHandler } from "./ErrorHandler";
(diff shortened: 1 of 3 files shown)


frontend/src/AppRouter.tsx
```diff
@@ -71,84 +72,92 @@ interface Props {
...
-                  <Switch>
-                    <Route
-                      path={r.routePath}
-                      component={r.component}
-                      render={r.render}
-                    />
-                  </Switch>
-                </Route>
-              ))}
+      <ErrorHandler>
```
syucream, line 75: Put it within `<Router>` because it cannot call `useHistory` from outside of the component.

hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for fixing its problem by this very elegant implementation 👍

PR #1015, opened 2023-12-02
Enhance test cases on entry form

syucream:
SSIA


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for making tests that are very good for example 👍

PR #1018, opened 2023-12-07
Change specification of Entry.get_available_attrs() for deleted referral for UX

userlocalhost:
This commit changed its specification not to show value that refers deleted Entry
at the array_named_object typed Attribute. This must increase for actual UX.


hinashi (approved):
LGTM!

PR #1021, opened 2023-12-12
Fixed update acl error when another role was already set

hinashi:
Steps to reproduce the bug

1. Create an entity
2. Only user A creates administrator role A
3. Set FULL permission only for role A on entity
4. Set the entity back to public
5. Change entity ACL settings with user B


userlocalhost (approved):
Thank you very much for fixing this hidden bug 👍

PR #1024, opened 2023-12-13
Split component in Entity and Entry List

hinashi:
There was a case where I wanted to use only the card part of the list, so I divided the components.


userlocalhost (approved):
Thank you for modularize each components 👍

PR #1025, opened 2023-12-13
Fixed advanced search when referral is deleted in array_named_entry type

hinashi:
Changed it to only delete the key if the entry is deleted.


userlocalhost (approved):

PR #1027, opened 2023-12-16
Enable to toggle webhook feature

syucream:
webhook is one of the rich feature provided by AirOne. I believe its not so useful for some users, so I'd like to enable/disable the feature, via environment variables.
If disabled,
- API v2 does not process creating/updating webhook settings
- Stop webhook requests
- Hide webhook form fields on the new UI


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for making this great work 👍

PR #1028, opened 2023-12-16
Notify field errors strongly

syucream:
Currently forms doesn't notify field level errors, show errors on each field alternatively. But I think its too silent. So I'd like to notify it via snackbar to provide stronger user feedback.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for improving UX 👍

PR #1029, opened 2023-12-18
Added an ID attribute to Entry exported yaml data for importing its data

userlocalhost:
This is necessary when import Entries using data that is exported from AirOne,
because previous implementation doesn't write Entry-ID attribute to export data.
Import processing couldn't understand which Entry is updated Entry or new created one.

This commit fixed that processing by adding Entry ID to export data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Added
 
 ### Changed
+* Added an ID attribute to Entry exported yaml data for importing its data.
+  Contributed by @userlocalhost
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


hinashi (approved):
LGTM!

I confirmed that the export data entry ID was added.
It seems that not using the entry ID when importing. Will this be addressed in another PR?

userlocalhost:
> It seems that not using the entry ID when importing. Will this be addressed in another PR?

Thank you very much for deeply checking this PR!!
I got it. I'll do it in this PR not to forget it.

PR #1032, opened 2023-12-24
Enhance FE test cases on entity domain

syucream:
SSIA


userlocalhost (approved):
Thank you very much for making tests that could be good example for making other test cases👍

PR #1033, opened 2023-12-26
Unified pagination footer by turning it into a component

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for all your hard work!!

PR #1043, opened 2024-01-06
Use useAsyncWithThrow basically instead of useAsync

syucream:
`useAsyncWithThrow()` custom hook throws an error so we can find wrong cases easily rather than `useAsync`.If it doesn't special requirements, lets uses `useAsyncWithThrow()` basically.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much to make it better 👍

PR #1044, opened 2024-01-08
Make common parts on some modals as a common component

syucream:
Most of modals have similar parts like title, description, caption, ... so I make them a common component.


hinashi (approved):
LGTM!

userlocalhost (approved):
How great work this is!! This must be necessary and make us alleviate in the future 👍 
★★★★★

PR #1045, opened 2024-01-09
Fixed an issue where array type attributes added later were not registered in ES

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for fixing a problem that is hard to find 👍

PR #1050, opened 2024-01-17
Fixed an issue where ACLs could not be read

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for fixing this very important problem 👍

PR #1051, opened 2024-01-18
Change not to override configuration parameter every time test instance is spawned

userlocalhost:
In the previous implementation of AironeTestCase, prefix word "test-" would be attached to the index name that is configured in the Django settings in the setUp() method of its class to distinguish elasticsearch index for test processing. But below exception woule be happend when there are many test cases in a repository because it would be chained every time test is run.
```
elasticsearch.exceptions.RequestError: RequestError(400, 'invalid_index_name_exception', 'Invalid index name [test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-test-airone], index name is too long, (256 > 255)')
```

This commit changes its re-configuration processing only one time.

### DIFF
--- a/airone/lib/test.py
+++ b/airone/lib/test.py
@@ -37,7 +37,10 @@ class AironeTestCase(TestCase):
 
     def setUp(self):
         OVERRIDE_ES_CONFIG = settings.ES_CONFIG.copy()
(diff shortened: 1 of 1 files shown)


hinashi (approved):
LGTM!

PR #1056, opened 2024-01-26
Changed so that SSO login is also redirected to the next URL

hinashi:
Since the next parameter was introduced, I changed it to use it.
https://github.com/python-social-auth/social-core/releases/tag/4.5.1


userlocalhost (approved):
Thank you very much for solving this problem 👍

PR #1057, opened 2024-01-27
Update mysql-innodboperator because helm-dependency-build fails

syucream:
Fix the build failures https://github.com/dmm-com/airone/actions/runs/7664204490/job/20888201486

(btw the operator seems that depends on MySQL 8.x so I wonder if it works fine without https://github.com/dmm-com/airone/pull/968#discussion_r1342084703


userlocalhost (approved):
Thank you for catch up with it 👍

hinashi (approved):
LGTM!

PR #1061, opened 2024-01-27
Enable collecting coverage on the frontend

syucream:
To visualize code coverage for the frontend with codecov, like the AirOne core.

### DIFF
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -70,3 +70,9 @@ jobs:
           NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
       - name: test
         run: npm run test
+      - name: codecov
+        uses: codecov/codecov-action@v3
+        with:
+          token: ${{secrets.CODECOV_TOKEN}}
+          flags: ui
+        continue-on-error: true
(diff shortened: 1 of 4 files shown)


.github/workflows/build-frontend.yml
```diff
@@ -70,3 +70,9 @@ jobs:
           NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
       - name: test
         run: npm run test
+      - name: codecov
+        uses: codecov/codecov-action@v3
+        with:
+          token: ${{secrets.CODECOV_TOKEN}}
```
syucream, line 76: It needs that owner's operation setting codecov token on GitHub Actions settings. Until the setting, it should fails so I set `continue-on-error: true` for now.

syucream: Hmm it possibly requires administrative operation on codecov side ...? 🤔 
https://github.com/dmm-com/airone/pull/1061#issuecomment-1913072630

userlocalhost (approved):
Thank you for updating it 👍 But I couldn't configure it by my authorization level ...
スクリーンショット 2024-01-29 15 47 19

hinashi (approved):
LGTM!

PR #1063, opened 2024-01-29
Added html id for uitest

hinashi:
SSIA


userlocalhost (approved):
Thank you for updating it 👍

PR #1068, opened 2024-02-04
Stop using fragments in menu

syucream:
To resolve the console error:

> Menu.js:162 MUI: The Menu component doesn't accept a Fragment as a child.
Consider providing an array instead.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for fixing FE implementation as proper manner.
My bad, I have frequently used `<>` as of alternative of `<Box>`. From now on, I'll never use `<>`.

PR #1072, opened 2024-02-08
Added processing to invoke TriggerAction when an Entry is created or updated through APIv1 and ordinary view

userlocalhost:
The new feature TriggerAction was added by https://github.com/dmm-com/airone/pull/1003.
But this feature is activated only from request via APIv2. This feature makes impact for users and they might require to use this from traditional UI (Django view and APIv1).

This PR exceptionally supports TriggerAction invocation processing to the traditional UI.

## ToDo
This PR should enable to invoke TriggerAction when user creates or updates Entry through
- [x] Django view processing (do_create and do_edit in the `entry/views.py`)
- [x] APIv1 processing
- [ ] import processing
- [ ] retrieve processing

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -1628,6 +1628,8 @@ class Entry(ACLBase):
             if attrv is None:
                 returning_attrs.append(
                     {
+                        "id": attr.id,
+                        "schema_id": attr.schema.id,
                         "name": attr.schema.name,
                         "value": AttributeValue.get_default_value(attr),
                     }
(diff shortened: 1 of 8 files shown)


entry/models.py
```diff
@@ -1628,6 +1628,8 @@ def to_dict(self, user, with_metainfo=False):
             if attrv is None:
                 returning_attrs.append(
                     {
+                        "id": attr.id,
+                        "schema_id": attr.schema.id,
```
userlocalhost, line 1632: ### Note
These might affect traditional API (`/api/v1/entry`) because it adds extra responses (`id` and `schema_id`).
But this makes it compatible with TriggerAction processing.
(c.f. https://github.com/dmm-com/airone/pull/1072/files#diff-e3b7f97a02a8c56838017005788cb61552a576c2ef1b8356528d807f8b89f482R39-R53)

hinashi (approved):
LGTM!

PR #1076, opened 2024-02-10
Try python match syntax to simplify type checking

syucream:
I propose a new coding style, using pattern matching introduced at Python 3.10:
https://qiita.com/ksato9700/items/3ce4c68c0d713874b693

We have a lot of matching logics, especially around checking attribute types. I think it helps to simplify the logics. It can replace `if attr_type == AttrType.STRING; elif attr_type == ...` pattern with `match attr_type; case AttrType.STRING; case ...`

### DIFF
--- a/airone/lib/elasticsearch.py
+++ b/airone/lib/elasticsearch.py
@@ -8,7 +8,7 @@ from elasticsearch import Elasticsearch
 
 from airone.lib.acl import ACLType
 from airone.lib.log import Logger
-from airone.lib.types import AttrTypeValue
+from airone.lib.types import AttrType, AttrTypeValue
 from entity.models import Entity
 from entry.settings import CONFIG
 from user.models import User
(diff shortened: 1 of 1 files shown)


airone/lib/elasticsearch.py
```diff
@@ -1030,61 +1028,67 @@ def make_search_results(
...
-                }
+            try:
+                attr_type = AttrType(attrinfo["type"])
+            except ValueError:
+                # For compatibility; continue that, and record the error
+                Logger.error("Invalid attribute type: %s" % attrinfo["type"])
+                continue
+
+            match attr_type:
+                case AttrType.STRING | AttrType.TEXT | AttrType.BOOLEAN:
```
syucream, line 1039: Its my sub proposal; how about using **enum based matching** instead of `== AttrTypeValue["..."]`?  The enum based approach is safer because the static check detects specifying wrong values, for e.g. `AttrType.INVALID` is disallowed by `mypy`

hinashi (approved):
LGTM!

userlocalhost (approved):
It's awesome!! That's what we need. Thank you very much for catching up with it and letting us know about this precious syntax 👍

PR #1077, opened 2024-02-11
Fix trigger form to make it suitable for RHF and zod

syucream:
- Replace `isSubmittable()`, a validation function with zod schema rules
  - Current implementation seems not to support creation case. It resolves the problem too
  - It can be more testable on zod schema's unit tests
- Fix some menu's to stop MUI warning logs

https://github.com/dmm-com/airone/assets/191684/192dbf09-564a-4690-ab57-fe81c1adaebf

### DIFF
--- a/frontend/src/components/trigger/ActionForm.tsx
+++ b/frontend/src/components/trigger/ActionForm.tsx
@@ -306,15 +306,14 @@ export const ActionForm: FC<Props> = ({
 
   return (
     <>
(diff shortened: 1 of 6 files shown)


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for fixing implementation problem that I made ･ﾟ･(｡✖д✖｡)･ﾟ･

PR #1079, opened 2024-02-11
Rename FE pages to unified naming convention

syucream:
Currently we have some page components freely named. It probably increases cognitive loads. So I would like to arrange the names to unified's, like `<domain, e.g. Entry><operation, e.g. Edit>Page.tsx`. In the rule, `EntryEditPage.tsx` is correct, `EditEntryPage.tsx` is bad.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for proposing good idea to increase readability 👍

PR #1080, opened 2024-02-12
Fix deps to avoid inifinite fetching

syucream:
Current deps on the `useEffect()` fetches entity/entry are bad. It occurs infinite fetching:
image


hinashi (approved):
LGTM!

userlocalhost (approved):
Wow, if this problem remained, it must incur incur user's wrath.
Thank you for fixing this important problem 👍 
You must save user's computers and our network resources!!

PR #1086, opened 2024-02-17
Replace with pattern match

syucream:
Make the type checking safer and simpler.


hinashi (approved):
LGTM!

userlocalhost (approved):
Good refactoring 👍 I'll use `match ~ case` syntax from now on.

PR #1090, opened 2024-02-22
Fixed invoke failure with do_edit

hinashi:
An exception error occurred depending on how do_edit was executed.


userlocalhost (approved):
Thank. you very much for prompt fixing 👍

PR #1093, opened 2024-02-23
Fix description for isTopLevel

syucream:
The new UI doesn't have side bar.


hinashi (approved):
LGTM!

userlocalhost (approved):
Good catch !! Thank you for fixing it 👍

PR #1095, opened 2024-02-27
Just to refactor for easily reducing copy and paste code and make it reusable from other processing

userlocalhost:
SSIA


hinashi (approved):
LGTM!

PR #1098, opened 2024-03-03
Improve job class with type hints and enum

syucream:
SSIA, to make it safer


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for its refactoring 👍

PR #1104, opened 2024-03-16
Make job status enum

syucream:
Enum is safe and smart rather than dict based static values.


userlocalhost (approved):
Thank you very much for refactoring them 👍 I'll use Enum instead of dict value.

hinashi (approved):
LGTM!

PR #1106, opened 2024-03-18
Added new feature that user can get referred Item's contents that an Item of advanced search result refers to from advanced search result page.

userlocalhost:
This introduces a new feature for advanced search result page that user can get Items contents that are referred from another Items in its page.

## Result of view page
(TBD)

## Rest of Work
- [x] API expansion
- [x] FE implementation


hinashi (approved):
LGTM!

PR #1108, opened 2024-03-19
Added type checking to trigger model

hinashi:
When using ForeignKey in a Django model, there was an issue where relation_name completion was not working.

Added type hints based on the issue below.
https://github.com/typeddjango/django-stubs/issues/579#issuecomment-950200788


userlocalhost (approved):
Thank you for improving 👍

PR #1110, opened 2024-03-22
Fixed entity editing failing for non created user

hinashi:
SSIA


syucream (approved):

userlocalhost (approved):
Thank you for fixing its processing 👍

PR #1114, opened 2024-03-26
Update DRF

syucream:
DRF has releases the latest version 3.15.x! Thats a new version since Sep 2022! 🎉 
https://www.django-rest-framework.org/community/release-notes/#315x-series

(and the version we currently use is terribly old ...) so I'd like to update it.


userlocalhost (approved):
Thank you very much for catching up with this news 👍 
It's a great propotion!

hinashi (approved):
LGTM!

PR #1115, opened 2024-03-26
Use django.conf.settings basically

syucream:
Which one is the best? I would like to change to use the best way if we have no reason.


hinashi (approved):
LGTM!

PR #1117, opened 2024-03-27
Remove redundant .value

syucream:
Enum values are directly usable! Calling `.value` is noisy and redundant, so I remove them to keep the code simple.

### DIFF
--- a/airone/lib/acl.py
+++ b/airone/lib/acl.py
@@ -10,7 +10,7 @@ class Iteratable(object):
         return self._types.__iter__()
 
 
-class ACLObjType(enum.Enum):
+class ACLObjType(enum.IntEnum):
     Entity = 1 << 0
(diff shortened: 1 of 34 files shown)


airone/lib/acl.py
```diff
@@ -10,7 +10,7 @@ def __iter__(self):
         return self._types.__iter__()
 
 
-class ACLObjType(enum.Enum):
+class ACLObjType(enum.IntEnum):
```
syucream, line 13: `IntEnum` ensures the values are `int`. It also can support implicit conversion on `any_value == SubClassOfIntEnum`

hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for blushing up implementation to be more maintainable.

PR #1119, opened 2024-03-29
Added env to add middleware settings

hinashi:
Added setting to add middleware settings in custom view.


userlocalhost (approved):
It's a hopeful change 👍

PR #1123, opened 2024-04-05
Fixed the index that adds trigger action

hinashi:
SSIA


userlocalhost (approved):

PR #1124, opened 2024-04-06
Simplify some remaining entry tasks additionally

syucream:
Simplify the most cases in celery tasks with the annotation.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for making it more simple.

PR #1126, opened 2024-04-07
Try test in parallel

syucream:
To accelerate Django testing, use [--parallel optiom](https://docs.djangoproject.com/en/5.0/topics/testing/overview/#speeding-up-the-tests) with following changes:

- Remove concurrent library probably unused (is is really okay?)
- Append pid suffix on ES index
- Append pid suffix on tmpfile dirname

As the result, its **2x faster** than current's.

before:
image

https://github.com/dmm-com/airone/actions/runs/8747605520/job/24006323011

after: 
image

https://github.com/dmm-com/airone/actions/runs/8768939258/job/24063893511?pr=1126

It also has a risk to make the testing flaky I guess. If we encounter some problems, I want to revert this once.


syucream:
Hmm its a bit faster but unstable ...

userlocalhost (approved):
Greaaaaaaat!! How good job 👍

hinashi (approved):
LGTM!

PR #1127, opened 2024-04-08
Fixed an error when specifying all entities in advanced search

hinashi:
Changed response format of entity attr API.


userlocalhost:
Thank you very much for fixing its critical problem promptly 👍 👍 👍

PR #1129, opened 2024-04-09
Fixed slow display of join_attr modal

hinashi:
Redrawing was suppressed by moving the component to a lower level.


userlocalhost (approved):
It's an excellent job!! Thank you very much for improving performance by this great refactoring 👍

PR #1130, opened 2024-04-11
Changed EntityAttrAPI response from name to index order

hinashi:
SSIA


userlocalhost (approved):
Thank you for fixing 👍

PR #1131, opened 2024-04-11
Fixed a problem not to be able to invoke Trigger when an Item is not updated

userlocalhost:
It is necessary to invoke trigger for applying setting default value trigger.


hinashi (approved):
LGTM!

PR #1134, opened 2024-04-13
Refactor entry exports with match statements

syucream:
SSIA

### DIFF
--- a/entity/api_v2/views.py
+++ b/entity/api_v2/views.py
@@ -297,7 +297,7 @@ class EntityAttrNameAPI(generics.GenericAPIView):
             return entity_attrs.values_list("name", flat=True).order_by("name").distinct()
         # single parent_entity case
         else:
-            return entity_attrs.values_list("name", flat=True).order_by("index")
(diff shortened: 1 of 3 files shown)


entity/api_v2/views.py
```diff
@@ -297,7 +297,7 @@ def get_queryset(self):
             return entity_attrs.values_list("name", flat=True).order_by("name").distinct()
         # single parent_entity case
         else:
-            return entity_attrs.values_list("name", flat=True).order_by("index")
+            return entity_attrs.values_list("name", flat=True).order_by("index", "pk")
```
syucream, line 300: Its a not core change but it fixes a flaky test case. The latest test result in the master branch has the same cause https://github.com/dmm-com/airone/actions/runs/8641820682/job/23691959470

hinashi (approved):
LGTM!

entity/api_v2/views.py
userlocalhost: Thank you very much for fixing it 👍

userlocalhost (approved):
Thank you very much for your great work every time 👍

PR #1136, opened 2024-04-13
Show 'ongoing' status on entity if it has non-finished jobs

syucream:
airone has many asynchronous jobs to process heavy data modifications. It means that users can wonder the data is the latest of not. (I guess some users doesn't care about the jobs usually)
So, as my proposal, it shows if  core data structure ( will be modified from jobs such `entity` ) has ongoing processes or not for usability.
image

How do you think about it?


userlocalhost:
I completely agree with you 👍

syucream:
I focus on entity create/edit operations in this PR. And also I separate lightweight and heavy processes on create/update entity in the API side.
How about this?

apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.0.14",
+  "version": "0.0.15",
```
syucream: Release the new version if no problem 🙇

PR #1146, opened 2024-04-28
Introduce devcontainer to get development environment more smoothly

syucream:
Try [devcontainer](https://code.visualstudio.com/docs/devcontainers/containers) as an alternative solution for https://github.com/dmm-com/airone/pull/935


syucream:
Hmm it mostly works but the frontend is still something wrong on runtime ...
I guess its happend by clsx version required by notistack but I have no more information ...

image

syucream:
After some tries, I become not to be able to reproduce https://github.com/dmm-com/airone/pull/1146#issuecomment-2094062755 ... maybe https://github.com/dmm-com/airone/pull/1146/commits/6b0d4f859f278cd3cae25dfa3d3d31cc55414617 resolve it. ( Give me feedback if you catch the same issue! )

Anyway I think its ready to use devcontainer to build development environment more smoothly! 🚀

.devcontainer/frontend/devcontainer.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "frontend",
```
syucream: Hmm I guess its better to aggregate frontend/backend containers to one container.

hinashi: Certainly, a single container is simpler and better.
I'll try using an Ubuntu image!

syucream: cool!

hinashi (approved):
LGTM!

PR #1148, opened 2024-04-28
Use enum more effectively

syucream:
Replace some self-defined iterable types or dict based types with enum with `@unique`, to make it safe and human-readable.


userlocalhost (approved):
Good work!! Thank you very much for utilizing such useful standard functions and making its implementation more robust 👍

hinashi (approved):
LGTM!

PR #1150, opened 2024-05-02
Added recurcive join attr function

hinashi:
SSIA


userlocalhost:
## Note
Fixed a problem that all default value of recursive join_attr would be "" was fixed by [here](https://github.com/dmm-com/pagoda/pull/1150/commits/91bae049175eafb1eabc2f48f705b39058f27fe8#diff-89f09bc74300c786f8ff965e14df7bfb6b9798b1a966b10ee6aef6422813ca57R322)

PR #1153, opened 2024-05-03
pydantic based type safe job params

syucream:
Current tasks handle untyped params via simply json string. To make it safer, I try define typed params with [pydantic](https://docs.pydantic.dev/latest/), a good way to handle typed data structure.

For now, I introduce that for entry import/export v2. I'd like to expand it for other jobs step by step.


hinashi (approved):
LGTM!

userlocalhost (approved):
Excellent!! It much safer than previous unstructured dict data and it must increase readability 👍 👍
Thank you very much for your great work 👍

PR #1158, opened 2024-05-11
Change the types Optional[T] to T | None mostly

syucream:
`Optional[T]` can be described with `T | None`, thats a smarter I think.
ref. https://stackoverflow.com/questions/69440494/python-3-10-optionaltype-or-type-none


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for letting me know about this clever solution and catching up with the state of art technology.

PR #1160, opened 2024-05-11
Typing exported entry attribute value

syucream:
SSIA, make it type safe more


userlocalhost (approved):
Thank you very much for making it more safe and maintainable 👍

hinashi (approved):
LGTM!

PR #1165, opened 2024-05-14
Added disabled to reset password

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for clewing it up 👍

PR #1169, opened 2024-05-18
Split testing in GHA to accelerate

syucream:
Split testing with matrix (see https://docs.github.com/ja/actions/using-jobs/using-a-matrix-for-your-jobs ) in GHA workflow to accelerate testing more.
image

before: 9m 31s
https://github.com/dmm-com/pagoda/actions/runs/9108967765

after: 6m 22s
https://github.com/dmm-com/pagoda/actions/runs/9140276412?pr=1169

But the splitting is complicated rather than one testing workflow. How do you think about this?

### DIFF
--- a/.github/workflows/build-core.yml
+++ b/.github/workflows/build-core.yml
@@ -9,7 +9,6 @@ on:
       - "**/*.py"
       - "templates/**/*.html"
       - ".github/workflows/build-core.yml"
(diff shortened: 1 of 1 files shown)


hinashi (approved):
LGTM!

userlocalhost (approved):
Excellent!! How great it is 👍

PR #1170, opened 2024-05-18
airone -> pagoda

syucream:
The docs page is now broken, probably because of baseURL.
https://dmm-com.github.io/pagoda/


userlocalhost (approved):
Good catch!! Thank you very much for fixing it 👍

hinashi (approved):
LGTM!

PR #1171, opened 2024-05-21
Fixed exception error in apiv2

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for fixing it promptly 👍

PR #1174, opened 2024-05-22
Added BASIC authentication log for debug

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for adapting this feature with very excellent solution 👍

PR #1175, opened 2024-05-23
Fixed failed generate_client.sh

hinashi:
SSIA


userlocalhost (approved):
How fast you find out this problem!! Thank you very much 👍

PR #1178, opened 2024-05-25
Migrate to enum based AttrType

syucream:
The enum based refer like `AttrType.KEY` is safer than `AttrTypeValue["key"]`, it can catch errors before runtime.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for your hard work 👍

PR #1180, opened 2024-05-27
Fixed LDAP authentication error

hinashi:
Fixed an error caused by https://github.com/dmm-com/pagoda/pull/1171


userlocalhost (approved):
Thank you very much for fixing it

PR #1181, opened 2024-05-27
Fixed failed export entry v2

hinashi:
SSIA


userlocalhost (approved):
It's great to find out this problem in advance 👍 👍

PR #1183, opened 2024-05-31
Changed custom view settings

hinashi:


userlocalhost (approved):
It's great!! Thank you very much for inventing this clever solution 👍

PR #1185, opened 2024-06-05
Fixed generate_client.sh error in docker

hinashi:
SSIA


userlocalhost (approved):
Thank you for blushing it up 👍

PR #1186, opened 2024-06-06
Removed Basic Auth support

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for shape it up 👍

PR #1188, opened 2024-06-08
Modernize attr types more

syucream:
Replacing some attr type values with enum based `AttrType`


userlocalhost (approved):
Thank you very much for your hard work and its good refactoring!! That makes its implementation more maintainable and it leads to make us motivated to do it.

hinashi (approved):
LGTM!

PR #1192, opened 2024-06-13
Added an option to be able to disable showing history page of Item from entry details page

userlocalhost:
This is a small change to be able to set administrator intentionally disable specific item of EntryControlMenu.


hinashi (approved):
LGTM!

PR #1195, opened 2024-06-15
pydantic models in es lib

syucream:
Enhance ES lib with pydantic


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for refactoring this implementation to robust it 👍

PR #1201, opened 2024-06-22
Make ACLType enum

syucream:
Migrate self-defined enum like type to python native enum type, to make it more type safe and simple.


hinashi (approved):
LGTM!

userlocalhost (approved):
Excellent!! Thank you very much for refactoring it 👍

PR #1202, opened 2024-06-22
Try to perform advanced search without ES

syucream:
Just a PoC that trying to remove ES on advanced search with acceptable performance.

My PoC action plans
- [ ] Preparation: Simply reimplement advanced search with MySQL
  - Do support: entry name filter, limit/offset pagination, some filter keys (contains, not contains, empty, not empty)
  - Don't support: including referrals, `is_output_all`, "duplicated" filter key, join_attr
- [ ] PlanA: Make some indexes for ^
- [ ] PlanB: Make a summary table that's a separated search index table to accelerate advanced search


syucream:
On very limited situation, I examined the performance of the rough PoC implementation with contains-keyword filter with 100+ entries.

Its not so bad performance ( currently, better than current ).
```
[Experimental Advanced Search] processing time: 0.152780 sec
[Current Advanced Search] processing time: 1.577253 sec
```

According to the django debug toolbar, its based on just 3 queries to `Entry`, `Attribute`, and `AttributeValue`. It looks like we have a chance to enhance the performance more with indexing on `ACLBase.name`

image

syucream:
I found current advanced search has a non-essential N+1 performance issue should be noise on this research. So I resolve it, on another PR: https://github.com/dmm-com/pagoda/pull/1213

syucream:
Currently I examine the PoC code with 10000+ entries:

```
mysql> select count(*) from entry_entry;
+----------+
| count(*) |
+----------+
|    10151 |
+----------+
1 row in set (0.00 sec)

mysql> select count(*) from entry_attribute;
+----------+
| count(*) |
+----------+
|   140552 |
+----------+
1 row in set (0.02 sec)

mysql> select count(*) from entry_attributevalue;
+----------+
| count(*) |
+----------+
|   230694 |
+----------+
1 row in set (0.03 sec)
```

And I found that fetching a lot of AttributeValue's makes the performance slow heavily ...
https://github.com/dmm-com/pagoda/pull/1202/files#diff-2ec0a0aa7fd7414b90b69d49c09c095635a38e9225a40b1825077dc567691900R52-R67

As the mitigation, I'll try some ideas:
- optimize the query more
- defer unnecessary fields
- query AttributeValue's concurrently
  - record-oriented: split records to chunks
  - attribute-oriented: split records by attribute

syucream:
I partially, simply perform them, with fetching widely without prefetch(will occur heavy query):

> optimize the query more
> defer unnecessary fields
> query AttributeValue's concurrently

Then the experimental advanced search performance was so so (**2x slower** than the current) ...

- 1st: current. the latency was 1.4s
- **2nd: experimental. the latency was 2.6s**
image

(I examined with the test dataset, ref. https://github.com/dmm-com/pagoda/pull/1202#issuecomment-2227226152 )

Now its available with `AIRONE_ENABLE_ESLESS_ADVANCED_SEARCH=true`!
It looks good, I will try another optimization including modifying exist models!

syucream:
https://github.com/dmm-com/pagoda/pull/1261 should be better solution, so I close this PR.

PR #1209, opened 2024-07-04
Added validate function for custom views

hinashi:
It was added because some entities may need to validate the name of the entry name.


userlocalhost (approved):
Thank you for introducing new feature to make its UX better 👍

PR #1212, opened 2024-07-06
Fix N+1 on entry copy

syucream:
Accelerate entry copy v2 API

# before / after

If triggering to copy 1000 entries ...

## before

Over 40s

image

## after

Faster!

image


userlocalhost (approved):
How awesome it is 🎉 !! Thank you very much for your excellent work!

hinashi (approved):
LGTM!

PR #1213, opened 2024-07-06
prefetch related sub models

syucream:
Remove N+1 on advanced search.

# before / after

Examined with 100+ entries have 4 attributes.

## before

About 6s, caused by heavy N+1.

image

image

## after

About 1s, with smart queries.

image

image


userlocalhost (approved):
Excellent 👍 Thank you very much for accelerate performance of this application.

hinashi (approved):
LGTM!

PR #1214, opened 2024-07-06
Add group/role relations

syucream:
Replace storing group/role id into `value` with new `group_id`, `role_id` columns.

Its possible to:
- make it safe and fast with foreign key constraint with index
- store values effectively because of string -> integer type changes
- perform prefetching to reduce N+1, like `attrv.all().select_related("group").select_related("role")`
- make the code readable more. (currently ... don't you think why it scans Group/Role by ".value"?

Now its just an idea, what do you think about this?

If the idea looks okay, we should take steps for safe:
1. update the newly written values
2. update existing values (data migration
3. use `group` and `role` on read workloads


syucream:
I've separated these parts as another PR.

> 1. update the newly written values
> 2. update existing values (data migration

https://github.com/dmm-com/pagoda/pull/1223

syucream:
In this PR, finally remove dependencies on .value for group, role types. Its a breaking change, apply https://github.com/dmm-com/pagoda/pull/1223 before this.

userlocalhost (approved):
Thank you very much for changing implementation to handle each Role and Group values at `.role` and `.group` of `AttributeValue` 👍

hinashi (approved):
LGTM!

PR #1216, opened 2024-07-09
Fixed some common components to be more reusability

userlocalhost:
This PR has following changes.

* To be able to hide cancel button at the <SubmitButton> component
* Apply for the case that server returns field-name as snake-case (e.g. `vlan_id`)
   - the `fieldRepoter()` method couldn't identify snake-case because zod schema is configured by camel case (like `vlanId`)


hinashi (approved):
LGTM!

PR #1217, opened 2024-07-10
Added tooltip for long entry name

hinashi:
On the entry list page, if the name was long, it would be truncated and not be visible in full.
A tooltip has been added.
image


userlocalhost (approved):
Good feature!! Thank you very much for solving that problem with reasonable idea 👍

PR #1220, opened 2024-07-11
Added safety processing to prevent changing user password when PASSWORD_RESET_DISABLED is set

userlocalhost:
In the current implementation certainly prevent to changing user password through UI. But, it's also necessary to doing it directly, that reject request its request on the server-side processing when is set at the settings.


hinashi (approved):
LGTM!

PR #1221, opened 2024-07-11
Added new middleware to prevent to exposing old URL when LEGACY_UI_DISABLED parameter is set in the settings.AIRONE

userlocalhost:
## Rest of work
- [x] write tests for new middleware GateKeeper


hinashi (approved):
LGTM!

PR #1223, opened 2024-07-14
Add group/role relations, and double write into them for data migration

syucream:
A part of https://github.com/dmm-com/pagoda/pull/1214 , so see the PR first of all.
It covers the steps:
1. update the newly written values
2. update existing values (data migration

For the 1st step, it appends `group`, `role` fields and stores these ids into the fields, with keeping existing logic.

For the 2nd step, we need to modify existing records to store `group_id` and `role_id`. For e.g. the following DML can perform it:
```sql
BEGIN;

UPDATE entry_attributevalue
SET
    group_id = CASE
        WHEN data_type IN (16, 1040) THEN CAST(value AS UNSIGNED)
        ELSE NULL
    END,
    role_id = CASE
        WHEN data_type IN (64, 1088) THEN CAST(value AS UNSIGNED)
        ELSE NULL
    END
WHERE
  -- group
  (data_type = 16)
  OR
  -- children of array_group
  (data_type = 1040 AND parent_attrv_id IS NOT NULL)
  OR
  -- role
  (data_type = 64)
  OR
  -- children of array_role
  (data_type = 1088 AND parent_attrv_id IS NOT NULL);

SELECT ROW_COUNT() AS updated_rows;

COMMIT;
```

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -56,6 +56,18 @@ class AttributeValue(models.Model):
     boolean = models.BooleanField(default=False)
     date = models.DateField(null=True)
     datetime = models.DateTimeField(null=True)
+    group = models.ForeignKey(
+        Group,
+        null=True,
+        related_name="referred_attr_value",
+        on_delete=models.SET_NULL,
+    )
+    role = models.ForeignKey(
+        Role,
+        null=True,
+        related_name="referred_attr_value",
+        on_delete=models.SET_NULL,
+    )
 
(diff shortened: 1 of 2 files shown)


userlocalhost (approved):
Thank you very much for dividing into small and easily understandable unit.
And I really appreciate to solve unreasonable implementation that handles all instance with `value` parameter, which is TextField.

hinashi (approved):
LGTM!

PR #1224, opened 2024-07-16
Fixed failed invoke trigger for oldUI

hinashi:
In the Old UI, entity_attr_id is always included, so we changed it to use that.


userlocalhost (approved):
Thank you very much for fixing it and blush up its implementation 👍

PR #1225, opened 2024-07-16
Added send mail when exception error in celery job

hinashi:
Changed the way celery handles exception errors.


userlocalhost (approved):
Thank you very much for making good feature to let us know a problem as fast as possible 👍

PR #1229, opened 2024-07-18
Fixed search_entries when specified multi entity

hinashi:
The problem occurred under the following conditions:
- is_output_all: True
- attrinfo: []
- entities: multiple entities specified


userlocalhost (approved):

PR #1232, opened 2024-07-20
Store prefetched entries in context to avoid N+1 on JobListAPI

syucream:
Accelerate job list API. Its **3.8x faster** than current implementation, 7s -> 1.8s!

image

### DIFF
--- a/job/api_v2/serializers.py
+++ b/job/api_v2/serializers.py
@@ -28,6 +28,8 @@ class JobSerializers(serializers.ModelSerializer):
     target = serializers.SerializerMethodField(method_name="get_target")
     passed_time = serializers.SerializerMethodField(method_name="get_passed_time")
 
+    PREFETCHED_ENTRIES_KEY = "__prefetched_entries"
+
     class Meta:
         model = Job
         fields = ["id", "text", "status", "operation", "created_at", "target", "passed_time"]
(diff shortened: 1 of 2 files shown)


userlocalhost (approved):
Excellent!! It's a great performance improvement 👍 
And thank you very much for telling us a good tip to do it by using `get_serlizer_context()`.

job/api_v2/serializers.py
```diff
@@ -36,12 +38,19 @@ class Meta:
     def get_target(self, obj: Job) -> JobTarget | None:
         if obj.target is not None:
             if obj.target.objtype == ACLObjType.Entry:
-                sub = Entry.objects.filter(id=obj.target.id).select_related("schema").first()
```
userlocalhost, line 39: memo:
This line refers each Entry foreach Entries, which are existed on the database because this `get_target()` method is called for each Entries. That leads to N+1 problems.

job/api_v2/serializers.py
```diff
@@ -36,12 +38,19 @@ class Meta:
...
+                sub: dict = self.context.get(self.PREFETCHED_ENTRIES_KEY, {}).get(obj.target.id)
```
userlocalhost, line 41: memo:
This `self.context` can get data that associated `get_serializer_context()` returns. This is useful to get whole data outside of this view.

hinashi (approved):
LGTM!

PR #1233, opened 2024-07-21
Enable target_id filter on the job list API v2

syucream:
SSIA, and its helpful to dig reason why target has ongoing changes.

https://github.com/user-attachments/assets/ae2ba913-0473-4cc5-ac4f-ef811a895a65


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.0.18",
+  "version": "0.0.19",
```
syucream, line 3: It needs library update 🙇

userlocalhost (approved):
Thank you very much for improving UX 👍

hinashi (approved):
LGTM!

PR #1235, opened 2024-07-26
Fix issue where re-added attribute's value appears in advanced search.

tsunoda-takahiro:
Resolve attribute display issue in advanced search after re-adding.


userlocalhost (approved):
LGTM 👍 Thank you for writing test to confirm fixing implementation would work expectedly.

hinashi (approved):
LGTM!

PR #1236, opened 2024-07-27
Add a testdata generator script to make development easily

syucream:
# Background

We sometimes implement/test logics with complicated data, for e.g. creating an entity/entry has all the type of attributes, via UI. But its usually painful.
And also we're performing a PoC https://github.com/dmm-com/pagoda/pull/1202 requires a lot of entries with complicated schema.

# Resolution

A testdata generator script makes Entity/EntityAttr/Entry/Attribute/AttributeValue with randomly generated payload.

## Easy to use

We just need to give num_of_entities, num_of_entries

```
# 2 entities, 10 entries on each entity
$ python tools/generate_testdata.py 2 10
Progress: entry creation 0/20
Progress: entry creation 1/20
...
Progress: entry creation 18/20
Progress: entry creation 19/20
Finished; generated test data timestamp: 20240728005654
```

Then you get generated testdata. NOTE it doesn't make ES index so we need to make it by ourselves if needed.

image
image

## Fast

Pagoda requires a lot of I/Os on creating a new entry that should be heavy. The script works in multi-threading so faster than just creating entries sequentially.

```
# A lot of entries ...
$ time python tools/generate_testdata.py 1 1000
...
Progress: entry creation 997/1000
Progress: entry creation 998/1000
Progress: entry creation 999/1000
Finished; generated test data timestamp: 20240728010916
python tools/generate_testdata.py 1 1000  90.57s user 14.77s system 112% cpu 1:33.94 total
```

## Randomly generated AttributeValues

It generates random values automatically, make testing complicated features, like advanced search, easy.

image

### DIFF
--- a/tools/generate_testdata.py
+++ b/tools/generate_testdata.py
@@ -1,0 +1,217 @@
+"""
+Test data generator, generates most of major model (roughly random) data in Pagoda, easily, fast.
+NOTE it doesn't make indexes for ES. Make it yourself if you need.
+
+How to use:
+$ python tools/generate_testdata.py [num_entities] [num_entries]
+- num_entities: The number of entities to generate.
+- num_entries: The number of entries to generate for each entity.
+"""
+
+import os
+import random
+import string
+import sys
+from concurrent.futures import Future, ThreadPoolExecutor
+from datetime import date, datetime, timezone
+from optparse import OptionParser
+
+import configurations
+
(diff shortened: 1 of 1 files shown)


userlocalhost (approved):
LGTM 👍

hinashi (approved):
LGTM!

PR #1237, opened 2024-07-28
Simplify attrv relations

syucream:
I guess some ManyToMany relations on Pagoda are redundant. How about simplifying it to 1:N relation? It makes relations simple, safe to keep that ( maintaining N->M, N<-M relations is so hard ) For now, I try to simplify parent - children AttributeValue relations.

NOTE it contains breaking changes on the DB schema and import export yaml format.


userlocalhost (approved):
How great it is!! I amazed that this PR doesn't require changing DB model and work it 👍

hinashi (approved):
LGTM!

PR #1238, opened 2024-07-30
Fixed title in login page

hinashi:
SSIA


userlocalhost (approved):
Thank you for fixing it 👍

PR #1242, opened 2024-08-03
Cleanup data_array.add() should be unnecessary

syucream:
It should be controlled by `parent_attrv` in `AttributeValue`


userlocalhost (approved):
Thank you very much for clearning implementation that is not necessary any more by changing its specification 👍

hinashi (approved):
LGTM!

PR #1244, opened 2024-08-03
Simplify Entry - Attribute relation 1:N

syucream:
Similar to https://github.com/dmm-com/pagoda/pull/1237, the relation looks redundant; in my understanding, single Attribute instance doesn't belong to multiple Entry instances.

### DIFF
--- a/entry/tests/test_model.py
+++ b/entry/tests/test_model.py
@@ -164,7 +164,6 @@ class ModelTest(AironeTestCase):
         entry.save()
 
         attr = self.make_attr("attr", entry=entry)
-        entry.attrs.add(attr)
 
         self.assertEqual(Entry.objects.count(), 2)
         self.assertEqual(Entry.objects.last().created_user, self._user)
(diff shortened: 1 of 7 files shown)


entry/tests/test_model.py
```diff
@@ -785,7 +779,7 @@ def test_coordinating_attribute_with_dynamically_added_one(self):
         # create new attributes which are appended after creation of Entity
         self._entry.complement_attrs(self._user)
 
-        self.assertEqual(self._entry.attrs.count(), 1)
+        self.assertEqual(self._entry.attrs.count(), 2)
```
syucream, line 782: `make_attr()` doesn't relate entry.attrs so currently `self._attr` isn't accesible via `self._entry.attrs` here. Is it intended? If not ... this PR resolves the problem, then the correct value should be changed to `2`.

hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for changing this great change that reduce redundant referral management 👍 And I'm very surprised that you can do it with very small change. How great Django ORMapping abstraction mechanism.

PR #1246, opened 2024-08-04
Refine Attribute.values relations

syucream:
WIP


syucream:
hmm Attribute <-> AttributeValue relation actually requires ManyToMany because Attribute wants only "parent" AttributeValue if the type is array-like.

PR #1247, opened 2024-08-05
Fixed an incorrect parameter in the SSO configuration

hinashi:
SSIA


userlocalhost (approved):

PR #1248, opened 2024-08-10
Fix entity - attrs relation

syucream:
SSIA, similar to https://github.com/dmm-com/pagoda/pull/1244


hinashi (approved):
LGTM!

userlocalhost (approved):
Wow!! Thank you very much for not only solving a traditional problem, but also settling its liabilities. I really appreciate your smart and dedicate work 🙏

PR #1249, opened 2024-08-10
Separate advanced search logics as a domain service

syucream:
Have a domain service executes domain logics separately from models, to simplify existing entry models.
With considering to layered architectures ref. https://qiita.com/ridai/items/009ed6e9ae9a4b02d879#onionarchitecture%E3%81%AB%E3%81%A4%E3%81%84%E3%81%A6


hinashi (approved):
LGTM!

userlocalhost (approved):
It's a good idea to separate application service of Entry from its model. Thank you for inventing a good multiple layer design in this application 👍

PR #1251, opened 2024-08-14
deleted users are not displayed

tsunoda-takahiro:
Fixed issue where deleted users were still displayed in the role management screen.

### DIFF
--- a/role/tests/test_api_v2.py
+++ b/role/tests/test_api_v2.py
@@ -359,6 +359,7 @@ class ViewTest(AironeViewTest):
         )
 
     def test_deleted_users_are_not_displayed(self):
+        self.admin_login()
         user = User.objects.create(username="test_user")
         role = Role.objects.create(name="test-role")
 
--- a/role/tests/test_api_v2.py
+++ b/role/tests/test_api_v2.py
@@ -359,7 +359,6 @@ class ViewTest(AironeViewTest):
         )
 
     def test_deleted_users_are_not_displayed(self):
-        admin = self.admin_login()
         user = User.objects.create(username="test_user")
         role = Role.objects.create(name="test-role")
 
--- a/role/tests/test_api_v2.py
+++ b/role/tests/test_api_v2.py
@@ -357,3 +357,22 @@ class ViewTest(AironeViewTest):
                 }
             ],
         )
+
+    def test_deleted_users_are_displayed(self):
+        admin = self.admin_login()
+        user: User = User.objects.create(username="test_user")
+        params = {
+            "name": "test-role",
+            "description": "test-description",
+            "users": [user.id],
+            "groups": [self.group1.id],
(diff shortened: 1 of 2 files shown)


role/tests/test_api_v2.py
```diff
@@ -357,3 +357,22 @@ def test_export(self):
                 }
             ],
         )
+
+    def test_deleted_users_are_displayed(self):
```
userlocalhost: It's better to name the one that is in accordance with this context. It ends up seeming that deleted user should be displayed from this test case name 😂

role/tests/test_api_v2.py
```diff
@@ -357,3 +357,22 @@ def test_export(self):
...
+            "description": "test-description",
+            "users": [user.id],
+            "groups": [self.group1.id],
+            "admin_users": [admin.id],
+            "admin_groups": [self.group2.id],
+        }
+        self.client.post("/role/api/v2/", json.dumps(params), "application/json")
+        self.client.delete("/user/api/v2/%d/" % user.id)
+        role = Role.objects.filter(name="test-role").first()
+        role_get = self.client.get(f"/role/api/v2/{role.id}")
```
userlocalhost: You had better to distinguish pre-processing and test processing that you want to test.

Most of tests (e.g. [test_create_entry_when_trigger_is_set](https://github.com/dmm-com/pagoda/blob/e43a16a67a6c2dfe6758833f9f6a6f3b77105a67/entity/tests/test_api_v2.py#L3575))  processing consists of following 3 steps.

1. Pre-processing. That makes data that is necessary for test processing
  - e.g. [Like here](https://github.com/dmm-com/pagoda/blob/e43a16a67a6c2dfe6758833f9f6a6f3b77105a67/entity/tests/test_api_v2.py#L3576-L3591)

2. Test processing. The main processing of test case what you want to check in the implementation. That would be very short processing in the API test because all you have to do is calling API with proper parameters. 
  - e.g. [Like here](https://github.com/dmm-com/pagoda/blob/e43a16a67a6c2dfe6758833f9f6a6f3b77105a67/entity/tests/test_api_v2.py#L3593-L3600)

3. Post-processing. After the main processing, you should check data context (and others) that main processing world as you expected.
  - e.g. [Like here](https://github.com/dmm-com/pagoda/blob/e43a16a67a6c2dfe6758833f9f6a6f3b77105a67/entity/tests/test_api_v2.py#L3601-L3608)

userlocalhost (commented):
Thank you very much for fixing its problem! It's OK to be merged. But if you change the things that I made comment, your code would be the one that looks like a professional 👍 

And I would appreciate it if you could fix its implementation to pass all CI processing before requesting review.

hinashi (approved):
LGTM!

userlocalhost (approved):
It's a good job 👍

PR #1252, opened 2024-08-16
Added dnd-kit library

hinashi:
Because I want to use dnd-kit libraries in my custom views.
https://dndkit.com/


userlocalhost (approved):
LGTM 👍

PR #1255, opened 2024-08-16
Fixed failure to detect AttrType

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for fixing it and your hard work 👍

PR #1256, opened 2024-08-16
Modernize entry models

syucream:
- Enhance type hints
- replace if-else with match-case


userlocalhost (approved):
It's good refactoring 👍 Thank you very much for your hard work!!

hinashi (approved):
LGTM!

PR #1257, opened 2024-08-17
Wrap all the tasks with the decorators

syucream:
Wrap all the tasks with our task decorators. No task resolve job_id themselves!

### DIFF
--- a/airone/lib/job.py
+++ b/airone/lib/job.py
@@ -9,27 +9,31 @@ from airone.lib.log import Logger
 from job.models import Job, JobStatus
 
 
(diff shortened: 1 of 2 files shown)


userlocalhost (approved):
It's a good change! It's very smart to separate following processing from task main processing.

1. checking processing whether task is ready to run.
2. post processing that set proper status for each jobs.

This change must increase maintainability and accelerate development 👍 
Thank you very much for its excellent work 💯

hinashi (approved):
LGTM!

PR #1260, opened 2024-08-21
bugfix deleted objects are not displayed.

tsunoda-takahiro:
Fixed a bug that caused deleted users and groups to remain visible in the role management screen.


hinashi (approved):
LGTM!

PR #1261, opened 2024-08-24
Put advanced search index in MySQL

syucream:
# Overview

As an alternative way to realize the advanced search feature, having search index in MySQL simply.
Its a PoC continued from https://github.com/dmm-com/pagoda/pull/1202

- Currently supported
  - basic search condition, entity & entities attrs 
  - basic attr filters, empty, not-empty, contains, not-contais
  - entry name filter
- NOT supported
  - including referrals
  - select from all entities
  - DUPLICATED attr filter
  - care for join_attrs, search chain
  - some special symbols (implicitly) control behavior of the search 

# Performance

With 10000+ entries, 14 attrs in an entry, the advanced search E2E latency is **about 2.5s**

image

The details of the search query to DB. Basically it uses index effectively. The core search scan is so slow, we can speed up if any improve ideas for that.

image

(However, current ES based search for the same data is faster, **1.3s** )

# How to use

Its enabled by default in this PR! If hope to disable, set `AIRONE_ENABLE_ESLESS_ADVANCED_SEARCH=false`

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -2293,14 +2293,12 @@ class AdvancedSearchAttributeIndex(models.Model):
     entity = models.ForeignKey(Entity, on_delete=models.CASCADE)
     entity_attr = models.ForeignKey(EntityAttr, on_delete=models.CASCADE)
     entry = models.ForeignKey(Entry, on_delete=models.CASCADE)
-    attr = models.ForeignKey(Attribute, on_delete=models.CASCADE)
+    attr = models.ForeignKey(Attribute, on_delete=models.CASCADE, null=True)
 
     type = models.IntegerField()
     entry_name = models.CharField(max_length=200)
     key = models.CharField(max_length=1024, null=True)
     raw_value = models.JSONField(null=True)
-
-    # TODO probably should have entry name
 
     class Meta:
         indexes = [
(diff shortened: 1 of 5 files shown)


entry/models.py
```diff
@@ -2287,3 +2287,127 @@ def _get_value(attrname: str, attrtype: int, value):
...
+
+class AdvancedSearchAttributeIndex(models.Model):
+    entity = models.ForeignKey(Entity, on_delete=models.CASCADE)
+    entity_attr = models.ForeignKey(EntityAttr, on_delete=models.CASCADE)
+    entry = models.ForeignKey(Entry, on_delete=models.CASCADE)
+    attr = models.ForeignKey(Attribute, on_delete=models.CASCADE, null=True)
+
+    type = models.IntegerField()
+    entry_name = models.CharField(max_length=200)
+    key = models.CharField(max_length=512, null=True)
```
syucream, line 2300: It might be problem, the max length is 512 for now because of the limit of MySQL index max length

hinashi: Textarea attribute and large amount of array data may exceed max.

userlocalhost:
Thank you very much for making a great implementation. Would you please making a simple test case to use this new feature to understand how to use it. Thank you in advanced.

syucream:
It remains this problem https://github.com/dmm-com/pagoda/pull/1261#discussion_r1744869854 but its a hard so I think its better to aim it on another PR.

@userlocalhost @hinashi How about this?

userlocalhost (approved):
Thank you very much for making it great progress for simplifying and rationalizing 👍

hinashi (approved):
LGTM!

PR #1262, opened 2024-08-25
Send date filteer in ISO8601

syucream:
Its better to handle date values in ISO format, not our original format, to improve interoperability on client-server.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for its good suggestion 👍

PR #1264, opened 2024-08-30
Added a helper components that is often used to adapt flex style

userlocalhost:
In many React components, we were made many StyledBox according to the purpose. But most of them are just a `Box` that `display:flex` is applied. 

This commit commoditize them as `FlexBox` and made derived ones `RightAlignedBox` and `CenterAlignedBox`. They might go on increasing, but I made simple cases so far.


hinashi (approved):
LGTM!

PR #1266, opened 2024-08-31
Use datepicker in datetime type too

syucream:
Just apply the date picker in the advanced search page to datetime attr filter.

image

Its NOT date TIME picker, excludes time(HH:mm:ss) part because most of users don't hope to filter with such exact condition, in my expectation.


userlocalhost (approved):
Thank you for handling DATETIME attribute type 👍

hinashi (approved):
LGTM!

PR #1267, opened 2024-08-31
Separate loading stype between normal and join_attr

syucream:
Without join_attrs, advanced search result page doesn't show loading icon on changing pages. Its not so kindful so I separate the loading style between these features, and ensure to show loading in the normal(default) advanced search.


userlocalhost (approved):
Thank you for caring join_attr feature and fixing it 👍

hinashi (approved):
LGTM!

PR #1269, opened 2024-09-03
Added exact option in custom route settings

hinashi:
If there is a route in the entry list in the custom route, the route for the entry details will be overwritten.

e.g. `entities/9999/entries/entries/1111/details`

custom_entries_path: `entities/${custom_entityId}/entries` ←catch
entry_details_path: `entities/${custom_entityId}/entries/${entryId}/details`


userlocalhost (approved):
It's good catch!! Thank you very much for fixing it 👍

PR #1272, opened 2024-09-07
Improve join attrs

syucream:
- Show loading icon on join attrs
 - simplify refreshing states/handlers


userlocalhost (approved):
Thank you very much for refactoring it 👍

PR #1273, opened 2024-09-07
Reduce N+1 on entry get APIv2

syucream:
It'll accelerate 3.2s -> 2.0s


userlocalhost (approved):
Thank you very much for improving performance and telling us how to reduce SQL by using select_related() method.

hinashi (approved):
LGTM!

PR #1278, opened 2024-09-14
Roughly support referrals search

syucream:
SSIA, to try implementing similar features with the current advanced search.
NOTE now it cannot filter out "entries" with the referral filter conditions because the referrals are determined "AFTER" the limit/offset pagination executed.

image

### DIFF
--- a/entry/services.py
+++ b/entry/services.py
@@ -9,6 +9,7 @@ from airone.lib.acl import ACLType
 from airone.lib.elasticsearch import (
     ESS,
     AdvancedSearchResultRecord,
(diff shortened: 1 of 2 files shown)


syucream:
https://github.com/dmm-com/pagoda/actions/runs/10859696623/job/30139379192?pr=1278

This CI failure should be fixed with https://github.com/dmm-com/pagoda/pull/1279 , maybe ...

userlocalhost (approved):
Thank you very much for implementing this necessary feature 👍

PR #1279, opened 2024-09-14
Fix webhook url verification cases

syucream:
Make the response success intentionally on the test cases, because probably the `POST http://example.net/` now returns not 2XX response.
ref. https://github.com/dmm-com/pagoda/actions/runs/10859696623/job/30139379192?pr=1278


userlocalhost (approved):
I'm embarrassed that I made a test case to send an actual request. Thank you very much for fixing them with very reasonable implementation 👍

hinashi (approved):
LGTM!

PR #1282, opened 2024-09-18
Added workaround for over key size case

hinashi:
This is temporary workaround.


syucream (approved):
thanks! and it'll better to have better solution for that as we talked before 🙌

userlocalhost (approved):
Thank you very much for fixing very important work 👍

PR #1283, opened 2024-09-22
Enable to import sjis encoded files

syucream:
`FileReader.readAsText()` encodes in utf-8 by default without 2nd argument `encoding?`, so it'll post invalid encoded files to import API. The export APIs will generate shift-jis encoded files so the problem is actually possible to occur.

So I introduce to detect the encoding before reading the content with [encoding-japanese](https://github.com/polygonplanet/encoding.js)


hinashi (approved):
LGTM!

userlocalhost (approved):

PR #1286, opened 2024-09-23
Avoid flicking on attr condition field

syucream:
The `AutocompleteWithAllSelector` component for attribute condition flicks on changes on entry field, caused by:
```
          {!attrs.loading && (
            <AutocompleteWithAllSelector
...
```

To avoid that, it just switch disabled state instead of switch show/hide state.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for this fixing. This must increase UX 👍

PR #1294, opened 2024-10-06
Refine AttrHint type

syucream:
Make `AttrHint` type more type-safe with pydantic model


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for making this implementation better than before 👍 I'm very happy it is modernized.

PR #1292, opened 2024-10-05
Update openapi-generator

syucream:
Follow recent changes on openapi-generator

PR #1296, opened 2024-10-08
Revert "Update openapi-generator"

userlocalhost:
This revert #1292 because following problem was emerged in some custom application after running `npm run watch:custom` as below.
https://github.com/OpenAPITools/openapi-generator/issues/14022

```
ERROR in ./frontend/src/apiclient/autogenerated/models/GoogleOAuthLoginUpdate.ts:15:10
TS2305: Module '"../runtime"' has no exported member 'exists'.
    13 |  */
    14 |
  > 15 | import { exists, mapValues } from "../runtime";
       |          ^^^^^^
    16 | /**
    17 |  *
    18 |  * @export

ERROR in ./frontend/src/apiclient/autogenerated/models/IpaddrIpaddrSchema.ts:15:10
TS2305: Module '"../runtime"' has no exported member 'exists'.
    13 |  */
    14 |
  > 15 | import { exists, mapValues } from "../runtime";
       |          ^^^^^^
    16 | /**
    17 |  *
    18 |  * @export
```
These files are generated by open-api generator. Those `exists` function doesn't exits in the `runtime` file and never be used, so we hope it won't be imported.

The serializer code that this error was happened is here.
```
class IpEntitySerializer(serializers.ModelSerializer):                                                                 
    class Meta:
        model = Entity                                                                                                 
        fields = ["id", "name"]  
```

We puzzled... :(

Notice: @syucream


hinashi:
This issue was caused by files generated by an older version of the generator remaining.
Resolved by deleting the `frontend/src/apiclient/autogenerated` directory and re-running the command.

PR #1299, opened 2024-10-12
Allow to change header color to show environment deferences etc

syucream:
SSIA, to identify the pagoda environment if we have some pagoda instances.

For e.g. with `AIRONE_HEADER_COLOR=red`:
image

The main use case is to show the environment. Today tech company serves multiple environments lilke development, qa, staging, production, etc. Its easy to show the difference, e.g. with `qa=default`, `staging=yellow`, `production=red`


hinashi (approved):
LGTM!

userlocalhost (approved):
Goood!! That is the thing that we want! Thank you very much 👍

PR #1300, opened 2024-10-13
Rename object-like attr types

syucream:
Rename display attr types from `*entry` to `*object` to suite the legacy UI
image


hinashi (approved):
LGTM!

userlocalhost (approved):
it's good idea to unify this type name. Good catch and I appreciate for your good work 👍

PR #1302, opened 2024-10-18
Changed the order of before delete custom views

hinashi:
Before the change, the call was made within background processing,
so even if the delete process was canceled in customview,
API request would return an OK response.


userlocalhost (approved):
Thank you very much for your work 👍 It looks OK.

PR #1303, opened 2024-10-19
Stop to show 1password shortcut

syucream:
Disable to show unnecessary 1password icons, with using `data-1p-ignore` attribute
https://developer.1password.com/docs/web/compatible-website-design/


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for increasing UX feature 👍

PR #1306, opened 2024-10-20
Use act() appropriately

syucream:
Use `act()` in the test codes to reduce the warning log:
```
    console.error
      Warning: An update to ReferralsAutocomplete inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
```


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for fixing this implementation 👍

PR #1310, opened 2024-10-30
Changed the way to specify custom routes

hinashi:
Due to changes in react-router specifications, Changed the way to specify custom routes.


userlocalhost (approved):
Thank you very much for adapting affects of react-router upgrading 👍

PR #1311, opened 2024-10-31
Fixed trigger form where value could not be entered

hinashi:
SSIA


userlocalhost (approved):
Great !! Thank you very much for your hard-work 👍

PR #1312, opened 2024-11-01
Keep to use our own ErrorBoundary

syucream:
react router v6 has the specific default ErrorBoundary in the module side, and its called before our's. We prefer to be called our one, so I insert a bridge to re-throw errors as an ErrorBoundary to react router.


hinashi (approved):
LGTM!

PR #1317, opened 2024-11-07
Changed to not update ElasticSearch when editing entity

hinashi:
If you have a large number of entries, updating ElasticSearch takes time.
If you many times edit entities, the load will increase.
We recommend not updating ElasticSearch when editing entities,
but rather updating ElasticSearch in a batch process once a day.


userlocalhost (approved):
Good!! Thank you very much for fixing its very important problem that would affect this application safety operation 👍

PR #1320, opened 2024-11-09
Refine dependencies for frontend

syucream:
Most (for now, all) of dependencies don't have concrete reasons why it must be in `dependencies`, not `devDependencies` because the frontend SPA expects to be pre-built, so I simplify the dependencies, with considering to build our own modules like https://github.com/dmm-com/pagoda/pull/1313


hinashi (approved):
LGTM!

userlocalhost (approved):
LGTM. Thank you very much for considering to make it better 👍

PR #1321, opened 2024-11-12
Introduce isReadonly flag at the AdvancedSearchResult page

userlocalhost:
This flag is used for hiding control components because this might be used at other pages to show its result as more simple style as below.
スクリーンショット 2024-11-12 12 15 26


hinashi (approved):
LGTM!

PR #1323, opened 2024-11-15
Introduced isDisabled at the AttributeValueField components and its sub-ordinate ones

userlocalhost:
The `isDisabled` parameter is humble, not be forward usually.
When an attribute want to be prohibited to change but just showing, this parameter is useful for its use-case like this.
スクリーンショット 2024-11-15 11 52 24


hinashi (approved):
LGTM!

PR #1326, opened 2024-11-18
Hidden Group and Role entries from the job list

tsunoda-takahiro:
Implemented a fix to hide new tasks for Group and Role from the job list.


hinashi (approved):
LGTM!

PR #1327, opened 2024-11-18
Enable to pass external parameters to React-UI implementation via Django template

userlocalhost:
This commit introduced new parameter `extendedGeneralParameters` that is expected to be used for general use-cases especially for specifying some value to custom-view.


hinashi (approved):
LGTM!

PR #1330, opened 2024-11-20
Changed to allow specifying limit of job api

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for its enhancement 👍

PR #1338, opened 2024-12-07
Remove dependency on react router from error handler

syucream:
`useNavigate()` should be called within react router `Router` context, but the `ErrorHandler` will not follow. So I remove dependency on the library from the component.

Currently the component doesn't work correctly:
```
router.js:284 Uncaught Error: useNavigate() may be used only in the context of a <Router> component.
    at invariant (router.js:284:11)
    at useNavigateUnstable (index.js:262:102)
    at useNavigate (index.js:259:46)
    at GenericError (ErrorHandler.tsx:64:55)
```


hinashi (approved):
LGTM!

userlocalhost (approved):

PR #1339, opened 2024-12-09
Fixed problem that might raise ElasticsearchException depends on context by SearchChain processing

userlocalhost:
But this is a workaround of it.


hinashi (approved):
LGTM!

PR #1340, opened 2024-12-10
Fixed not being able to cancel a job from the Job menu

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for enhancing its feature to be more useful 👍

PR #1341, opened 2024-12-11
Changed git command for display the version

hinashi:
It was excluded because it would be confused with the pagoda-core tag.


userlocalhost (approved):
Thank you very much for solving it 👍

PR #1350, opened 2025-01-08
Fixed response type of the Entry API

hinashi:
There was an issue where some API responses did not have the alias attribute.


userlocalhost (approved):

PR #1352, opened 2025-01-12
Make notistack externals to allow custom views to use/customize the notifications

syucream:
`pagoda-core` module fixes notistack and its context so custom views cannot reuse that appropriately


hinashi (approved):
LGTM!

userlocalhost (approved):

PR #1355, opened 2025-01-20
Enable to get items considering with Alias name

userlocalhost:
This commit update /entity/api/v2 API handler to be able to get item considering with related item.

## Confirmation
I confirmed that I can get item that is related with specified alias at the alias list page.
スクリーンショット 2025-01-20 18 18 36


entity/api_v2/views.py
```diff
@@ -192,11 +192,29 @@ def destroy(self, request: Request, *args, **kwargs) -> Response:
...
+        # filtered by specified "search" parameter
+        if request.query_params.get("with_alias"):
+            return original_fields + ["aliases__name"]
+        else:
+            return original_fields
+
+
+@extend_schema(
+    parameters=[
+        OpenApiParameter("with_alias", OpenApiTypes.STR, OpenApiParameter.QUERY),
```
userlocalhost, line 210: It would be better to use `OpenApiTypes.BOOL` instead of `OpenApiTypes.STR`

userlocalhost: On second thought, `OpenApiTypes.STR` is better than `BOOL` because this is a GET parameter.

The `request.query_params.get("with_alias")` contains `string` typed value even through `extend_schema` configuration.
```Python
        if request.query_params.get("with_alias"):
            return original_fields + ["aliases__name"]
        else:
            return original_fields
```

(c.f. https://github.com/dmm-com/pagoda/pull/1355/files#diff-1322718f3873c6b8312fafb3c1eab3172964dd9a3eeeacd3ce3e5f39fadb01bfR202)

So that condition would be `True` when user specified `with_alias=false`.

hinashi (approved):
LGTM!

PR #1356, opened 2025-01-22
Optimize role import for job execution

tsunoda-takahiro:
The role import functionality has been modified to run as a background process due to its long processing time.


hinashi (approved):
LGTM!

PR #1358, opened 2025-01-23
Fix filtering of inactive users and groups in RoleExportAPI

tsunoda-takahiro:
This PR fixes the issue where inactive users and groups were included in the RoleExportAPI response. Filters have been applied to ensure only active data is exported.


hinashi (approved):
LGTM!

PR #1367, opened 2025-02-13
Fixed edit category acl setting

hinashi:
An exception error occurred when setting the ACL for a category.


userlocalhost (approved):

PR #1369, opened 2025-02-14
Changed link style overall

hinashi:
Changed link color and underline for AirOne original.

https://hamidasimono.com/archives/2290


userlocalhost (approved):
Thank you very much for your prompt work and your hard work 👍

PR #1370, opened 2025-02-15
Upgrade to MUI v6

syucream:
- Upgrade MUI packages to version 6.x
- Update Emotion packages to latest versions
- Modify date picker components to use new prop syntax
- Replace deprecated props with new slotProps
- Update package-lock.json with new dependency versions


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/pagoda-core",
-  "version": "1.0.4",
+  "version": "1.1.0",
```
syucream, line 3: MUI upgrade should be a big change so I think its good to bump up the version.

userlocalhost (approved):
Thank you very much for its important work 👍

hinashi (approved):
LGTM!

PR #1371, opened 2025-02-16
Upgrade python

syucream:
SSIA


entity/api_v2/views.py
```diff
@@ -32,6 +30,23 @@
 from user.models import History, User
 
 
+# distutils.util.strtoboolの代替実装
+def strtobool(val):
```
userlocalhost, line 34: Thank you very much for supporting backward compatibility 👍

userlocalhost (approved):
Good!! Thank you very much for your great work 👍

hinashi (approved):
LGTM!

PR #1375, opened 2025-02-20
Added all select button in advanced search page

hinashi:
Added for removing bulk items.


userlocalhost:
## Discussion (Worring)
Current implementation delete whole Items of specified Model, whatever filter conditions are set.

But It might against user expectation when user set some filter conditions like below.

* Model: Test2
* Attr: label
* condition: `label=hoge`
スクリーンショット 2025-03-04 17 07 45

When user who expects to delete all Items that are matched with specified condition pushes this `YES` button, he(or she) must be confused and anger because whole Items are deleted.

### PlanA: disable isAll when user set configuration
This is a compromise solution

### PlanB: delete all Items that are matched with user specified filter configuration
This must be better than PlanA for usability and matched  with UX.

userlocalhost:
> PlanB: delete all Items that are matched with user specified filter configuration

This feature has equipped by 464a9cf9e602ce9a73e91dafb198279e23dd8086 commit.

userlocalhost (approved):
Thank you for your cooperation, @hinashi 🙌 
Now, this has already been ready for be merged, I think 😄

userlocalhost:
We have to care about the following cases.
- [x] Fixing description of alert box when filter condition is set.
- [x] It fails to delete when join_attr(s) are set.
  スクリーンショット 2025-03-05 10 27 49

userlocalhost:
I fixed Dialog description to change its context depends on its situation.

### 1. No filter conditions are set
スクリーンショット 2025-03-06 16 27 22

### 2. Filter conditions are set
スクリーンショット 2025-03-06 16 28 05

### 3. Join-attr conditions are set
スクリーンショット 2025-03-06 16 26 52

PR #1378, opened 2025-03-01
Update npm dependencies to latest versions

syucream:
SSIA


userlocalhost (approved):
Thank you very much for keeping dependencies updated not only for Python's but also frontend's ones 👍

hinashi (approved):
LGTM!

PR #1381, opened 2025-03-02
Fix some page components recently added

syucream:
- add page component snapshot tests
- rename them to follow naming convension (should be `XXXListPage`)


hinashi (approved):
LGTM!

userlocalhost (approved):
Awesome!! Thank you very much for its important work 👍

PR #1383, opened 2025-03-05
Restored simple search results page

hinashi:
There was a problem with simple search due to the following PR.
https://github.com/dmm-com/pagoda/pull/1368


userlocalhost (approved):
Excellent!! Thank you very much for your prompt work 👍

PR #1385, opened 2025-03-08
feat: Prevent 4-byte character input in entry form validation

syucream:
Add validation to reject entries with 4-byte characters (emojis) in:
- Entry name
- String attribute values
- Array string attribute values
- Object name fields

image


hinashi (approved):
LGTM!

userlocalhost (approved):
How excellent it is!! Thank you very much for your great contribution 👍

PR #1389, opened 2025-03-10
Added parameter to make item name a UUID

hinashi:
Depending on the model, want to use a UUID for item name.
A parameter has been added that will automatically enter a UUID when creating an item.


userlocalhost (approved):

PR #1390, opened 2025-03-11
Fixed creating attribute would fail if exists deleted attribute

hinashi:
There are cases where an attribute has been deleted and remains.
In such cases, there was an issue where a new attribute could not be created.


userlocalhost (approved):

PR #1395, opened 2025-03-15
Fix TypeScript type safety issues and improve error handling

syucream:
Restrict not to use `any` in most cases.
Today, our development style is dramatically changing with gAI, and strict typing helps AI works. So I would like to improve typing for AI, not humans :)


userlocalhost (approved):
Excellent!! And thank you very much for fixing any type specifications that are mostly set temporarily by me 😭. I won't use `any` anymore!

hinashi (approved):
LGTM!

PR #1397, opened 2025-03-21
Fixed booleans of NAMED_OBJECT_BOOLEAN type is not displayed correctly

hinashi:
SSIA


userlocalhost (approved):
LGTM 👍 Thank you for fixing it!

PR #1404, opened 2025-03-25
Fixed failing refresh after filtering in advanced search

hinashi:
The refresh was being run twice so I removed it.


userlocalhost (approved):
Excellent!! Thank you very much for fixing this critical problem 👍

PR #1406, opened 2025-03-27
Fixed join attr error in advanced search

hinashi:
Fixed an error when excluding attributes in join_attr in advanced search.


userlocalhost (approved):
Excellent!! Thank you very much for fixing its critical problem 👍

PR #1407, opened 2025-03-27
Fixed behavior when there are no duplicate items in advanced search

hinashi:
There was an issue where filtering was not performed when there were no duplicate items.


userlocalhost (approved):
Good hack!! Thank you very much for fixing with elegant code.

PR #1408, opened 2025-03-29
validate if entity has referrals

syucream:
Deny empty referral for a object-like attribute
image


hinashi (approved):
LGTM!

userlocalhost (approved):
Good feature!! Thank you very much for making this very useful function 👍

PR #1410, opened 2025-03-31
Fixed join_attr error after new attribute is added

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for fixing this important problem 👍

PR #1411, opened 2025-04-03
Fixed error with multiple entities in search_chain

hinashi:
When multiple entities were targeted in a search_chain,
an exception error occurred if the attribute did not exist in one of the entities.


userlocalhost (approved):
Good work! Thank you very much for fixing it 👍

PR #1412, opened 2025-04-06
Add error handling for non-UTF8 encodings in file download response

syucream:
Some encodings will fail at unsupported characters. It allows to continue encoding with replacing that to `?`.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for fixing its encoding problem with small change 👍

PR #1413, opened 2025-04-08
Fix entry copy error caused by empty or whitespace-only lines

tsunoda-takahiro:
This PR fixes an issue where submitting entry names with empty or whitespace-only lines caused a 400 error from the backend API during copy operations.
To resolve this, entry names are now trimmed and filtered before the API request is made.


hinashi (approved):
LGTM!

PR #1414, opened 2025-04-08
Fix layout overlap issue on small screens

tsunoda-takahiro:
Restructured layout to prevent UI elements from overlapping on smaller screens.
Applied responsive design for user panel display.


hinashi (approved):
LGTM!

PR #1416, opened 2025-04-09
Fixed validate referral in edit entity

hinashi:
In the case of the array attribute, unintended validation occurred.


userlocalhost (approved):
Good catch! And thank you for fixing its problem 👍

PR #1417, opened 2025-04-09
Fixed an error with invalid attrinfo parameter in entry searchAPI

hinashi:
SSIA


userlocalhost (approved):
Good work 👍 Thank you for fixing it.

PR #1418, opened 2025-04-09
Enhance RoleAPI with filter and ordering support

tsunoda-takahiro:
Improved RoleAPI by enabling ordering and filtering capabilities via DRF's built-in backends.


hinashi (approved):
LGTM!

PR #1419, opened 2025-04-10
Add ordering support to group tree API

tsunoda-takahiro:
Enabled alphabetical ordering of parent groups in the /groups/tree API response.


hinashi (approved):
LGTM!

PR #1420, opened 2025-04-11
Fix: Hide group links and menu for unauthorized users

tsunoda-takahiro:
Restricted access to group detail pages and actions from users without sufficient permissions.


hinashi (approved):
LGTM!

PR #1421, opened 2025-04-12
Enhance API response types with Pydantic

syucream:
Define strict request/response models to ensure type safety in API v1.


userlocalhost (approved):
Nice work 👍 Thank you very much for maintaining its traditional code.

PR #1423, opened 2025-04-18
Restrict access to RoleEditPage for unauthorized users

tsunoda-takahiro:
Adds frontend check using isEditable to prevent users without permission from viewing the role edit page. Shows a 403 error when accessed via direct URL.


hinashi (approved):
LGTM!

PR #1424, opened 2025-04-18
Improve: Group member filtering supports mixed case input

tsunoda-takahiro:
Refactored user filter logic to ignore case when filtering group members.


hinashi (approved):
LGTM!

PR #1425, opened 2025-04-19
Initialize array values in entry form fields and remove useEffect hooks

syucream:
The fields for array string and array named object will have multiple initial values accidentally with multiple renderings. The basic scenario is using React strict mode, so I believe its basically no problem, but its better to ensure the components are acceptable for multiple renderings.


hinashi (approved):
LGTM!

PR #1426, opened 2025-04-20
Support a rich entry name filter similar to attrs

syucream:
SSIA

image


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.8.0",
+  "version": "0.9.0",
```
syucream, line 3: needs to publish 🙏

hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for its important work and your hard work 👍

PR #1427, opened 2025-04-22
Fix permission-based UI rendering in UserList

tsunoda-takahiro:
Restricts access to edit-related UI components in UserList depending on user role.


hinashi (approved):
LGTM!

PR #1430, opened 2025-04-25
Fix breadcrumb rendering in group edit page

tsunoda-takahiro:
Fixes breadcrumb label to show "create" or "edit" depending on groupId. Snapshot updated. Deprecated act usage remains out of scope.


hinashi (approved):
LGTM!

PR #1431, opened 2025-05-02
Added helper processings to retrieve referral items that specified Item refers

userlocalhost:
I was mad whenever I wrote processing that refers Items from specified ones using Entry.get_attrv().

Because we have to refer multiple items `Entry -> AttributeValue -> ACLBase`, then convert its type from `ACLBase` to `Entry` like this.
```Python
entry = Entry.objects.get(...)
attrv = item.get_attrv("attribute name")
if attrv and attrv.referral and attrv.referral.is_active:
  target_entry = attrv.referral.entry
```

This get Item (`Entry`) that is referred by another Item (type `Entry`) directly as below.
```Python
entry = Entry.objects.get(...)
target_entry = entry.get_attrv_item("attribute name")
```

I believe this alleviate developer to write these code.

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -1,7 +1,7 @@
 import re
 from collections.abc import Iterable
 from datetime import date, datetime
-from typing import Any, List, Optional, Type
+from typing import Any, List, Optional, Type, Union
 
 from django.conf import settings
 from django.db import models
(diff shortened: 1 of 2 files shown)


hinashi (approved):
LGTM!

PR #1436, opened 2025-05-06
Allows the APIv2 advanced search to match entries without specified attributes if it has no attribute filter condition

syucream:
SSIA, search entries in multiple entities more flexibly:

image


hinashi (approved):
LGTM!

userlocalhost (approved):
How good work! Thank you very much for solving its problem and your hard work 👍

PR #1438, opened 2025-05-13
Introduce new feature to refer referral items from an Entry instance using Prefeth()

userlocalhost:
## Background
When user develops CustomView processing, user must handles built-in Item structure.

## Problem
In that case, user can choose following method to get attribute value or another Item that target Item's attribute value has.
```Python
from entry.models import Entry

item = Entry.objects.get(id=ITEM_ID)

# get attribute value of target attribute
item.get_attrv(attr_name).value

# get another Item that target attribute refers
item.get_attrv_item(attr_name).referral.entry

# refer A.attr1 -> B.attr2 -> C
A.get_attrv_item("attr1").referral.entry.get_attrv_item("attr2").referral.entry
```

However, both processing issues multiple SQLs internally. Therefore, it  may lead to N+1 problem.
To prevent it, Django has [prefetch_related()](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#prefetch-related) processing that reduce issuing multiple SQLs by using JOIN internally.

But Pagoda doesn't prepare standard interface to use it. So it's required much effort to implement it.

## What this PR
This PR introduces new feature `ItemWalker` that provides developer to implement above processing using Prefetch with unified way like this.
```Python
from entry.models import Entry, ItemWalker

iw = ItemWalker(ITEM_ID, {"attr1": {"attr2": {}}})

# refer A.attr1 -> B.attr2 -> C
iw["attr1"]["attr2"].item
```

This achieve both well performance (less issuing SQLs) and making code simpler :)


hinashi (approved):
LGTM!

PR #1439, opened 2025-05-13
Fix tokenLifetime validation and display unlimited case correctly

tsunoda-takahiro:
- Adjusted Zod schema to coerce string to number for tokenLifetime
- Unified validation error message to match legacy UI
- Clarified UI message for when tokenLifetime is set to 0 (unlimited)


hinashi (approved):
LGTM!

PR #1444, opened 2025-05-17
refactor: consolidate page and query state management in usePage hook

syucream:
- Aggregate query filter into `usePage`
- Perform URL encoding on the query to avoid accidental errors


userlocalhost (approved):
How great refactoring 👍 Thank you very much for your dedicate work !!

hinashi (approved):
LGTM!

PR #1446, opened 2025-05-18
Replace predefined method_table with register decorator to avoid cycle imports

syucream:
Simplify module dependencies, previously job <-> tasks requires each others, now just tasks require the job.


userlocalhost (approved):
How great it is!! It's quite obvious what event handler associated with what Job. Thank you very much for making this genius commit.

hinashi (approved):
LGTM!

PR #1449, opened 2025-05-20
Add real-time API search to role selector

tsunoda-takahiro:
RoleAttributeValueField now fetches role options dynamically as the user types, improving UX and scalability.

### DIFF
--- a/frontend/src/repository/AironeApiClient.ts
+++ b/frontend/src/repository/AironeApiClient.ts
@@ -584,6 +584,13 @@ class AironeApiClient {
     return await this.role.roleApiV2Retrieve({ id: roleId });
   }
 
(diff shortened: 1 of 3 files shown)


frontend/src/repository/AironeApiClient.ts
```diff
@@ -584,6 +584,13 @@ class AironeApiClient {
     return await this.role.roleApiV2Retrieve({ id: roleId });
   }
 
+  async searchRoles(keyword?: string): Promise<Role[]> {
```
hinashi: Since there is already a function called getRoles that uses the same API,
I thought it would be better to extend it.

hinashi (approved):
LGTM!

PR #1453, opened 2025-05-24
fix: handle empty string case in Elasticsearch regex pattern generation

syucream:
Ensure not to raise `IndexError` on the pattern builder. Currently its possible that a user filter advanced search results with spaces like ` `, `　`


hinashi (approved):

userlocalhost (approved):
Excellent both guys 👍 Thank you very much!!

PR #1454, opened 2025-05-24
Fix entry filter `clear` button

syucream:
It accidentally required double clicks to clear selected conditions, so I fix it.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for fixing it promptly 👍

PR #1455, opened 2025-05-26
Fix layout and wrapping for entry history values

tsunoda-takahiro:
Applied max-width and text wrapping to prevent layout breakage in EntryHistoryList.


hinashi (approved):
LGTM!

PR #1456, opened 2025-05-26
Fixed an issue where the snackbar would disappear in restore entry

hinashi:
SSIA


userlocalhost (approved):

PR #1457, opened 2025-05-26
Reorder attribute type dropdown and handle missing definitions

tsunoda-takahiro:
Reordered the attribute type dropdown list as per expected UX order. Guarded against undefined keys to ensure safe rendering.


hinashi (approved):
LGTM!

PR #1458, opened 2025-05-27
Fix missing snackbar on superuser password reset

tsunoda-takahiro:
Ensure snackbar appears before navigation by delegating responsibility to parent component.


hinashi:
I have confirmed that this is fixed when editing passwords from the user edit page.
The same problem occurred when editing passwords from the user list page.

hinashi (approved):
LGTM!

PR #1460, opened 2025-05-31
chore: migrate from codecov to octocov for test coverage reporting

syucream:
Use [octocov](https://github.com/k1LoW/octocov) instead of codecov. Now codecov doesn't work well...


userlocalhost (approved):
It's great to reduce depending third party system. Thank you very much for introducing octocov. 👍

hinashi (approved):
LGTM!

PR #1462, opened 2025-06-04
Fix unexpected modal closing on Tab key

tsunoda-takahiro:
Resolved an issue where the password edit modal would close when pressing the Tab key. Moved modal outside of UserControlMenu to avoid unmount on menu close.


hinashi (approved):
LGTM!

PR #1463, opened 2025-06-04
Added new parameter "exclude_referrals" and "include_referrals" for advanced_search

userlocalhost:
The new parameter `exclude_referrals` is used for excluding items that are referred by Items, which are belonged to specified model, from the result of advanced search. That is useful to narrow down speific Items for some use-cases (e.g. picking up items that are referred by multiple items).

And this also adds `include_referrals` parameter, which is opposite of `exclude_referrals`, to filter Items that are referred from specified Model's Item.


hinashi (approved):
LGTM!

PR #1464, opened 2025-06-05
Prevent self-deletion of user accounts

tsunoda-takahiro:
This PR disables the delete option for the current user and adds an extra safeguard against accidental self-deletion.


hinashi (approved):
LGTM!

PR #1469, opened 2025-06-07
refactor: move attribute name duplication check from service to schema validation

syucream:
Migrate service layer logic into zod schema validation


hinashi (approved):
LGTM!

userlocalhost (approved):
It's a nice measure to decouple policy and mechanism 👍 Thank you very much for this great refactoring. This must increase maintainability and helps future developers!

PR #1473, opened 2025-06-14
enhance type hints

syucream:
Make it more type-safe!


userlocalhost (approved):
Thank you very much for enhancing its type hint.

PR #1477, opened 2025-06-14
Fix broken octocov steps

syucream:
It will resolve the broken(silent) workflow ...


userlocalhost (approved):

hinashi (approved):
LGTM!

PR #1479, opened 2025-06-18
Enable to select display alternative attribute name at table head title by attrFilter configuration

userlocalhost:
This feature is useful when you want to display more simple attribute name rather than actual one using Search Results component.


hinashi (approved):
LGTM!

PR #1481, opened 2025-06-20
Added new parameter "omitHeadline" to the SearchResults component

userlocalhost:
This new parameter enables modify first column to edit icon button instead of item name.


hinashi (approved):
LGTM!

PR #1486, opened 2025-06-23
Prevent eager attr_referrals fetch on entry edit

tsunoda-takahiro:
Only fetch referrals after dropdown is opened or user types.


hinashi (approved):
LGTM!

PR #1487, opened 2025-06-23
Fixed an issue where the value of the array_string attribute of ItemWalker could not be obtained

hinashi:
Added support for `array_string` and `boolean` in ItemWalker.


userlocalhost (approved):
Excellent!! Thank you very much for its expanding 👍

PR #1488, opened 2025-06-26
Add dynamic document titles to main pages using usePageTitle

tsunoda-takahiro:
This PR introduces the usePageTitle hook to several main pages (Entry, Entity, Role, User, etc.) to provide dynamic and consistent document titles.


hinashi (approved):
LGTM!

PR #1490, opened 2025-06-27
Added UsePageTitle hook to export list

hinashi:
ref: https://github.com/dmm-com/pagoda/pull/1488


userlocalhost (approved):
Thank you for your assistance 👍

PR #1491, opened 2025-06-28
Support default values in some entity attr types

syucream:
Enable entities to have attribute default values:

image

Then it applies the values to a new entry:

image

Now the default values are only for some primitive types STRING, TEXT, BOOLEAN, NUMBER.


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.13.0",
+  "version": "0.14.0",
```
syucream, line 3: needs to publish 🙇

hinashi:
```
--
-- Add field default_value to entityattr
--
ALTER TABLE `entity_entityattr` ADD COLUMN `default_value` json NULL;
--
-- Add field default_value to historicalentityattr
--
ALTER TABLE `entity_historicalentityattr` ADD COLUMN `default_value` json NULL;
```

hinashi (approved):
LGTM!

userlocalhost (approved):
How great work it is!! Thank you very much for making its very important feature promptly 👍

PR #1493, opened 2025-07-05
Support array number type

syucream:
Support array number type, like array string

image


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.14.0",
+  "version": "0.15.0",
```
syucream, line 3: needs to publish~

userlocalhost (approved):
@syucream Thank you very much for implementing it.

And thank you for finishing up for making it. @hinashi

PR #1495, opened 2025-07-09
Added isNarrowDown parameters to the advanced search results page

hinashi:
Added parameters to the advanced search results page.

The following parameters are available:
- isReadonly: Whether or not there is a delete function.
- isNarrowDown: Whether or not there is a narrow down function.
- omitHeadline: Whether or not there is a display item name


userlocalhost (approved):
Thank you very much for enhancing its feature to be able to increase reusability 👍

PR #1498, opened 2025-07-19
Fix to restore named object's

syucream:
Restoring some attribute values is broken because of lack of appropriate value conversions.


hinashi (approved):
LGTM!

userlocalhost (approved):
Good work 👍 Thank you very much for fixing this important feature!

PR #1508, opened 2025-08-02
refactor: remove unnecessary React imports from component files

syucream:
Use `react-jsx`.

ref. https://zenn.dev/mh4gf/scraps/fcaab0c7f23e59


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for introducing new feature `ReactNode` that seems like more flexible type than FC.

PR #1510, opened 2025-08-07
Added ssl enable parameter

hinashi:
Changed to allow switching via environment variables since login fails in environments that do not use SSL


userlocalhost (approved):
LGTM 👍

PR #1511, opened 2025-08-09
Replace poetry with uv

syucream:
`uv` is a de facto, faster module management tools, a good replacement of `poetry`


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you very much for replacing modernized one 👍

PR #1512, opened 2025-08-10
Support entry-'itself' history

syucream:
Include entry 'itself' history on the entry history page.

image


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.15.0",
+  "version": "0.16.0",
```
syucream, line 3: need to publish 🚀

hinashi (approved):
LGTM!

userlocalhost (approved):
Gooood👍 Finally, it is made!!
Thank you very much. This must be the one that user have been longed.

PR #1513, opened 2025-08-16
Build pagoda-core module with production mode

syucream:
Currently the module `pagoda-core` published with development mode. But generally extended application with the module is built with production mode finally.


hinashi (approved):
Thanks for fixing this!

PR #1516, opened 2025-08-18
Fixed not being redirected after copying an item

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for doing this important work 👍

PR #1517, opened 2025-08-19
Added new parameter 'hint_item_name' to filter search result by item's name at the EntrySearchChainSerializer

userlocalhost:
In the previous implementation there is no way to filter result items by name at the EntrySearchChainSerializer.search_entries(). This PR implemented its feature by introducing new parameter "hint_item_name" that will be passed at the `hint_entry` parameter of AdvancedSearchService.search_entries().


hinashi (approved):
LGTM!

PR #1519, opened 2025-08-23
Stop to show the prompt on submit

syucream:
It also shows the prompt on successful submit, its so noisy.


hinashi (approved):
LGTM!

userlocalhost (approved):
Thank you for making its improvement to increase UX 👍

PR #1521, opened 2025-08-25
Changed to uv in clear and initdb script

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for updating and maintaining it 👍

PR #1523, opened 2025-08-27
An example code for applying custom_view (DO NOT MERGE)

userlocalhost:
SSIA


userlocalhost:
@syucream has already been challenged this feature. So it should be closed.

PR #1528, opened 2025-09-04
Added is_unmatch option in trigger conditions

hinashi:
```
ALTER TABLE `trigger_triggercondition` ADD COLUMN `is_unmatch` bool DEFAULT b'0' NOT NULL;
ALTER TABLE `trigger_triggercondition` ALTER COLUMN `is_unmatch` DROP DEFAULT;
```


userlocalhost (approved):
Thank you very much for making this important work 👍

PR #1531, opened 2025-09-10
Changed to send an empty array if no value is set for array attribute

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for adding processing to filter empty data to prevent adding unnecessary ArrayNamedObject attribute value.

PR #1532, opened 2025-09-11
Fix pagination on Advanced Search results

tsunoda-takahiro:
Pass totalCount to PaginationFooter (instead of count) and set siblingCount=1 to show adjacent pages. Prevents missing page 4 and neighbors. Files: SearchResults.tsx, PaginationFooter.tsx.


hinashi (approved):
LGTM!

PR #1533, opened 2025-09-14
Support backend plugin system roughly

syucream:
Similar pluggable backend architecture to https://github.com/dmm-com/pagoda/pull/1525

Pagoda plugins can implement extensible features with the SDK provides some pre-exported core APIs, and can be installed via python package management systems like pip.

- Extensible, flexible rather than the custom view
- Able to co-operate with python ecosystem
- Enable to have custom API endpoints, entity/entry hooks

Plugin developers can have outside of the project, like https://github.com/syucream/pagoda-backend-plugin-example as an simple plugin example.


userlocalhost (approved):
Thank you very much for doing this great work 👍

PR #1534, opened 2025-09-16
Bugfix: Pagination stability for Advanced Search

tsunoda-takahiro:
- Prevent duplicate requests on page change
- Ignore late responses from previous page
- Minimal, page-only change; no server changes


hinashi (approved):
LGTM!

PR #1540, opened 2025-10-27
Fixed empty values ​​could not be registered with the array_named_object_boolean attribute

hinashi:
SSIA


userlocalhost (approved):

PR #1541, opened 2025-10-28
This supports AttributeValue.date and .datetime for ItemWalker

userlocalhost:
Current ItemWalker doesn't support AttributeValue.date and .datetime parameters. Then, this PR do it.


hinashi (approved):
LGTM!

PR #1550, opened 2025-11-16
Add Pydantic models for enhanced validation in serializers

syucream:
Use pydantic to make APIv2 more type safe

### DIFF
--- a/entity/api_v2/serializers.py
+++ b/entity/api_v2/serializers.py
@@ -8,6 +8,8 @@ import requests
 from django.conf import settings
 from django.core.validators import URLValidator
 from drf_spectacular.utils import extend_schema_field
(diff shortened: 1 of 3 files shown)


userlocalhost:
Thank you very much for implementing it using Pydantic instead of current `WebhookSerializer. _validate_default_value_for_type()` that doesn't check input value considering its data structure.

I have a question about this PR.
I guess this PR is (almost) refactoring current validation processing of Entity(and Entry) api_v2 validation using Pydantic.
But I don't know the input value that you want to prevent to accepting.
Would you please tell me (or write an ideal test case) that you want to do?

userlocalhost (approved):
Thank you very much for improving serializer processing using Pydantic feature 👍

PR #1554, opened 2025-11-19
Added model referral filter to advanced search page

hinashi:
close: https://github.com/dmm-com/airone/issues/3126

Advanced search now allows to filter referrers by model.
image


userlocalhost (approved):
Thank you very much for making user required feature 👍

PR #1555, opened 2025-11-19
Fixed error in redirect long character URL

hinashi:
close: https://github.com/dmm-com/airone/issues/3275

Errors due to the following changes
https://github.com/django/django/commit/c880530ddd4fabd5939bab0e148bebe36699432a


userlocalhost (approved):
Good!! Thank you for fixing it 👍

PR #1559, opened 2025-11-22
JobAPI: Add support for DELETE_ENTITY_V2 and DELETE_ENTRY_V2 operations

syucream:
Always show entity/entry deletion jobs because users currently cannot confirm the deletion operations on the job list.


hinashi (approved):
LGTM!

PR #1560, opened 2025-11-23
Fix mypy type errors and improve type annotations across multiple files

syucream:
Current pagoda typing is not so well-safe because some checks are omitted. I'd like to enhance it to keep type safety continuously.


hinashi (approved):
LGTM!

PR #1563, opened 2025-11-29
Fix mypy type errors and improve type annotations in multiple files

syucream:
continued from https://github.com/dmm-com/pagoda/pull/1560 , It ensures type safety more!


userlocalhost (approved):
Good work 👍 Thank you very much for fixing it!

PR #1565, opened 2025-11-29
Use pydantic models in entity job tasks to ensure job params type safety

syucream:
Current job params don't have any explicit schemas. Making pydantic models enforces the mechanism to eusure type safety -- like _editing entity params must have id, name, type, ..._

### DIFF
--- a/entity/tasks.py
+++ b/entity/tasks.py
@@ -1,13 +1,249 @@
-import json
+from typing import Optional
+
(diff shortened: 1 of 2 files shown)


userlocalhost (approved):
Excellent!! How great this is! This must increase maintainability and save future working cost. Thank you very much for your great contribution 👍 👍

PR #1569, opened 2025-12-06
AdvancedSearchResultsPage: add hintEntry to export job registration

syucream:
Currently the exported data contains filtered entries; the entry name filter is just ignored. Its unexpected behavior. I'd like to fix it.


hinashi (approved):
LGTM!

PR #1571, opened 2025-12-13
Apply recent React security updates

syucream:
Recently React has some security issues. `pagoda` doesn't have so much security risk, but should apply the updates.


hinashi (approved):
LGTM!

PR #1572, opened 2025-12-13
refactor: remove babel-plugin-transform-class-properties and related configurations

syucream:
`babel-plugin-transform-class-properties` is currently unused, and has some security issues.


hinashi (approved):
LGTM!

PR #1573, opened 2025-12-13
chore: update Django and related dependencies to latest versions

syucream:
Upgrade some modules to prepare Django 5.2.


hinashi (approved):
LGTM!

PR #1575, opened 2025-12-20
style: adjust table cell widths for better layout

syucream:
It takes sufficient width (17% -> 30%) to show long entry names more readable. 17% is too short.


hinashi (approved):
LGTM!

PR #1576, opened 2025-12-20
feat: implement entity history tracking with detailed change

syucream:
Implement TBD values to show change details appropriately in the entity history page. It uses django-simple-history to resolve before/after value changes.


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.18.0",
+  "version": "0.19.0",
```
syucream, line 3: Requires to update the client to contain changes field.

hinashi (approved):
LGTM!

PR #1579, opened 2025-12-28
Make useTypedParams safer

syucream:
Allow it to check if required parameters are served. `useParams` will return undefined values.


userlocalhost (approved):
Thank you very much for making a feature that react UI processing is more safer 👍

PR #1582, opened 2026-01-04
Improve frontend test coverage widely

syucream:
Current FE tests are still cheap (check if it throws). It improves the test coverage.


hinashi (approved):
LGTM!

PR #1583, opened 2026-01-06
Changed the button has been disabled if it is not available

hinashi:
The advanced search's bulk update function requires target model IDs,
so it cannot be used if the search scope is not narrowed or if JOIN_ATTR is used.
Therefore, Changed the button has been disabled.

Fixed a bug in bulk updates when multiple models were selected.

### DIFF
--- a/frontend/src/components/entry/AdvancedSearchEditModal.tsx
+++ b/frontend/src/components/entry/AdvancedSearchEditModal.tsx
@@ -1,3 +1,4 @@
+import { AttributeData } from "@dmm-com/airone-apiclient-typescript-fetch";
 import { zodResolver } from "@hookform/resolvers/zod";
 import { Box, Button } from "@mui/material";
 import { useSnackbar } from "notistack";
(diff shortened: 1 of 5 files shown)


frontend/src/components/entry/AdvancedSearchEditModal.tsx
```diff
@@ -86,6 +87,31 @@ export const AdvancedSearchEditModal: FC<Props> = ({
           // Close this modal
           handleClose();
         });
+    };
+    // The aironeApiClient.bulkUpdateEntries would be called only one time in most cases.
+    // because it's rare that multiple Models are specified for advanced search page.
+    modelIds.forEach((modelId: number) => {
+      // In case multiple Models are specified, fetch the attribute schema for each Model
```
userlocalhost, line 94: Thank you very much for adding description of this special case 👍

userlocalhost (approved):
Thank you very much for doing this very important feature to increase usability 👍

PR #1584, opened 2026-01-08
Added allow empty option of UseTypedParams

hinashi:
There was an issue where the model creation page and trigger creation page could not be opened.


userlocalhost (approved):
Thank you very much for fixing degradation 👍

PR #1586, opened 2026-01-11
Switch rendered menu based on user permissions

syucream:
Current menus for a category, entity, entry will show disallowed operations. For e.g. it shows `edit` item even if a user doesn't have sufficient permissions.

It'll confuse so I fix the menu items based on a login user permission.


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.19.0",
+  "version": "0.20.0",
```
syucream, line 3: requires publish

hinashi (approved):
LGTM!

PR #1590, opened 2026-01-16
Added feature to attach ID parameter at each exported items

userlocalhost:
Close: https://github.com/dmm-com/airone/issues/2404

This enables user to update item name by import processing.


hinashi (approved):
LGTM!

PR #1599, opened 2026-01-24
Prepare to upgrade elasticsearch to 8.x

syucream:
ES7 is no longer supported. ( ref. https://endoflife.date/elasticsearch ) And it will block upgrading Django.

ES language (python) client supports forward compatibility so we should upgrade ES server side first. This PR prepares for that.
https://github.com/elastic/elasticsearch-py?tab=readme-ov-file#compatibility


userlocalhost (approved):
Thank you very much for maintaining it 👍 It's a good work!

PR #1600, opened 2026-01-24
Enables plugins to validate if an entity follows expected schema

syucream:
https://github.com/dmm-com/pagoda/pull/1592 allows to customize entity-specialized pages with (probably) expected schemas. For e.g. A plugin shows rich views with fetching contents from `url` attribute. In this case, the plugin expects the entity has `url` attribute`. But actually the entity can breaks the structure.

Its an actual example schema; requires `hostname: string`. If an entity doesn't follow the schema, it shows the schema mismatch error page.

```javascript
const sampleAttrSchema = z.object({
  // Require "hostname" attribute with STRING type
  // This will fail for entities without this attribute
  hostname: z.object({
    type: z.literal(AttrType.STRING),
  }),
});
```


userlocalhost (approved):
Awesome!! Thank you very much for making a great feature 👍
And UI is very cool.
Image

hinashi (approved):
LGTM!

PR #1601, opened 2026-01-30
Fixed build failure when npm run link:client at the latest environment

userlocalhost:
Here is the execution result in my environment.
スクリーンショット 2026-01-30 17 12 46


hinashi (approved):
LGTM!

userlocalhost:
Sorry, I found a problem that updating and creating Model (and Item) processings would be failed even through server-side processings were passed correctly, like this.


https://github.com/user-attachments/assets/a6205676-75c8-430e-bbed-aecced885ad0

That is because current ambiguous statement makes apiclient-typefetch code that will return empty value that cause JSON parase error at frontend processing even through server processing was passed, as below.
```JavaScript
    async entityApiV2Update(requestParameters: EntityApiV2UpdateRequest, initOverrides?: RequestInit | runtime.InitOverrideFunction): Promise<EntityUpdate> {
        const response = await this.entityApiV2UpdateRaw(requestParameters, initOverrides);
        return await response.value();   // ← It's nonsense!
    }
```

PR #1609, opened 2026-02-11
Migrate to Elasticsearch8

syucream:
cont'd from https://github.com/dmm-com/pagoda/pull/1599, upgrade ES client version.


userlocalhost (approved):
Thank you very much for applying code to Elasticsearch8 👍

PR #1612, opened 2026-02-14
Use standard mock

syucream:
Use unittest.mock


userlocalhost (approved):
Thank you for refactoring it 👍

PR #1616, opened 2026-02-20
Support type 3081 in entity attr validation

tsunoda-takahiro:
Handle special case for ARRAY_NAMED_OBJECT_BOOLEAN type to maintain backward compatibility


hinashi (approved):
LGTM!

PR #1621, opened 2026-02-24
Replace prettier with biome

syucream:
Now [biome](https://biomejs.dev/) is better formatter than prettier.
ref. https://qiita.com/tseno/items/0a9f048333cb850e4a6d


userlocalhost (approved):
Thank you for selecting a better library for keeping good implementation 👍

hinashi (approved):
LGTM!

PR #1622, opened 2026-02-24
Update TypeScript configuration and ESLint settings to support ES2022 features

syucream:
Use newer build target on the FE components.


userlocalhost (commented):
Thank you very much for maintaining its configuration to keep up with newer ECMAScript 👍

userlocalhost (approved):
Thank you very much for maintaining its configuration to keep up with newer ECMAScript 👍

hinashi (approved):
LGTM!

PR #1623, opened 2026-02-24
Perfectly remove poetry

syucream:
We've already used `uv` instead of `poetry` but some docs still expect using `poetry`, so I'd like to remove them.


userlocalhost (approved):
Good👍 Thank you very much for replacing new method and maintaining documents.

hinashi (approved):
LGTM!

PR #1624, opened 2026-02-24
Upgrade RabbitMQ to version 4.1.3 in Docker and Helm configurations

syucream:
RabbitMQ 3.8.x is too old...
https://endoflife.date/rabbitmq


userlocalhost (approved):
Great! Thank you very much for upgrading its dependent system 👍

hinashi (approved):
LGTM!

PR #1639, opened 2026-03-07
replace django-replicated with django-multidb-router

syucream:
[django-replicated](https://github.com/yandex/django_replicated) is no longer maintained and archived. It will be an issue on upgrade Django in the future.
[django-multidb-router](https://github.com/jbalogh/django-multidb-router) looks like a good replacement.

About the deletion on `airone/middleware/db.py`. For read-after-write consistency, django-multidb-router supports to forward requests to master DB served by `PinningRouterMiddleware`. So we don't need to have our own custom middleware.

> PinningRouterMiddleware attaches a cookie to any user agent who has just written. The cookie should be set to expire at a time longer than your replication lag. By default, its value is a conservative 15 seconds, but it can be adjusted like so:

ref. https://github.com/jbalogh/django-multidb-router?tab=readme-ov-file#configuration


hinashi (approved):
LGTM!

PR #1644, opened 2026-03-14
docs: clarify interaction between hooks and overrides in plugin system

syucream:
The `hooks` and `overrides` mechanism on the BE plugin will conflict. It describe the problem.


hinashi (approved):
LGTM!

PR #1646, opened 2026-03-14
chore: update dependencies and remove deprecated packages in package.json

syucream:
Current FE involves some trivial dependencies. It increases complexity and bundle size. This PR eliminate them.


hinashi (approved):
LGTM!

PR #1652, opened 2026-03-20
feat(tests): integrate pytest for testing and add pytest configuration

syucream:
With using pytest, especially pytest-xdist, accelerate the testing.


syucream:
hmmm it badly effects on the CI...

PR #1655, opened 2026-03-21
Split large tests to accelerate testing lead time

syucream:
`uv run coverage run manage.py test --parallel` works based on testing classes. So splitting large classes will improve the lead time. (In the CI job, it already splits by modules so the improvement effect is not so worked, but will effect in our local envs)


userlocalhost (approved):
You have a good eye 👍 Thank you very much for this measure to contribute for saving operation cost!

hinashi (approved):
LGTM!

PR #1658, opened 2026-03-24
Replaced pkg_resources because deprecated

hinashi:
Since pkg_resources was removed from setuptools, it has been replaced by importlib.metadata.


tsunoda-takahiro:
LGTM!

PR #1663, opened 2026-03-28
Use default permission classes specified in the settings

syucream:
It has the default permission class, just use it.
https://github.com/dmm-com/pagoda/blob/d5317163450164b3b59a5b739298fb678c17fa09/airone/settings_common.py#L477-L479


userlocalhost (approved):
Good catch! Thank you very much for removing unnecessary code 👍

PR #1664, opened 2026-03-28
avoid to overwrite staticfiles settings

syucream:
Reassign to `STORAGES` deletes `staticfiles` values.


userlocalhost (approved):
Thank you for fixing a potential problem to remove other necessary configurations by overwriting whole STORAGES value 👍

PR #1670, opened 2026-04-04
fix: improve type hints across various modules and update mypy configuration

syucream:
make it more type safe.


hinashi (approved):
LGTM!

PR #1671, opened 2026-04-04
fix: add postinstall script and update airone-apiclient dependency to optional

syucream:
Probably GitHub Registry at dmm-com restricts clients in the outside of the org, so `npm install` will fails. The change has a chance to avoid the issue with generate-client script.


hinashi (approved):
As you mentioned, this organization restricts the source IP address now.

PR #1672, opened 2026-04-06
Fixed CSV header(at the fist line) to list attribute names by index order

userlocalhost:
## Execution Results
Here is the result of CSV export of a specific Model (Test02) when attribute index order was configured as below.
```CSV
Name,foo,bar
tmp01,hoge,20260202
tmp02,fuga,20210001
```
スクリーンショット 2026-04-06 10 50 32

Then, here is the result of CSV export after changing its configuration as below.
```CSV
Name,bar,foo
tmp01,20260202,hoge
tmp02,20210001,fuga
```
スクリーンショット 2026-04-06 11 10 21

These results show that the order of CSV header and contents are followed expectedly by changing its attribute index order.

### DIFF
--- a/entry/tasks.py
+++ b/entry/tasks.py
@@ -375,9 +375,11 @@ def _yaml_export_v2(
 @register_job_task(JobOperation.CREATE_ENTRY)
 @app.task(bind=True)
 @may_schedule_until_job_is_ready_with_handlers(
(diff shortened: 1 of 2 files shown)


hinashi (approved):
LGTM!

PR #1678, opened 2026-04-10
Changed to invoke trigger when referenced object entry is deleted

hinashi:
closed: https://github.com/dmm-com/airone/issues/3445

When an object-type attribute's referral is deleted (soft-delete), the
attribute value is effectively None. Add TriggerCondition.get_invoked_actions_on_delete()
to evaluate trigger conditions at deletion time, and hook it into the
delete_entry / delete_entry_v2 tasks.

ARRAY types correctly compute remaining elements after deletion to avoid
false positives when the array still has other valid references.

### DIFF
--- a/entry/tasks.py
+++ b/entry/tasks.py
@@ -53,6 +53,7 @@ from entry.services import AdvancedSearchService
 from group.models import Group
 from job.models import Job, JobOperation, JobStatus
 from role.models import Role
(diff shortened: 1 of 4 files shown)


userlocalhost (approved):
How fast it is 👍 Thank you very much for solving this important problem promptly!

PR #1681, opened 2026-04-11
fix: update endpoint filtering to include plugin endpoints in schema generation

syucream:
It resolves some errors on `npm run generate:client`


hinashi (approved):
LGTM!

PR #1684, opened 2026-04-12
feat: add pagoda-e2e-verify skill for browser-driven integration testing

syucream:
An agent skill verify actual application behaviors with browsing. It should help developers to improve UX.


hinashi (approved):
LGTM!

PR #1686, opened 2026-04-16
Fix apply_join_attrs dropping referred entries due to search limit

hinashi:
- Fix a bug in `apply_join_attrs` where referred entries were silently
  dropped when the referred entity had more active entries than the
  `limit=len(ref_ids_in_entity)+100` cap passed to `search_entries`
- Add `entry_ids` parameter to `search_entries` and `make_query` to
  restrict ES search to specific entry IDs via an `ids` query filter,
  eliminating the limit-based data loss
- Rewrite the fetch loop in `apply_join_attrs` to use `entry_ids` with
  a chunk size of 1000, guaranteeing all referred entries are returned
  regardless of the total entity size
- Remove unused parameters (`entry_limit`, `is_output_all`,
  `exclude_referrals`, `include_referrals`) from the `apply_join_attrs`
  signature and all call sites

### DIFF
--- a/airone/lib/elasticsearch.py
+++ b/airone/lib/elasticsearch.py
@@ -261,6 +261,7 @@ def make_query(
     allow_missing_attributes: bool = False,
     exclude_referrals: list[int] = [],
     include_referrals: list[int] = [],
+    entry_ids: list[int] | None = None,
 ) -> dict[str, Any]:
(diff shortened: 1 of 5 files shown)


userlocalhost (approved):
Excellent! Thank you very much for fixing this problem with excellent code 👍

PR #1687, opened 2026-04-16
Reject isolated entries when setting attribute values via API

hinashi:
Previously isolation only hid entries from the referral selection UI.
Direct API calls could still assign isolated entries as attribute values.

- Add entity_attr param to validate_attr_value() and check isolation in
  _is_validate_attr_object(); api_v2 create/update returns 400 on violation
- Add isolation check in Attribute.add_value() for OBJECT/NAMED_OBJECT types
  to cover api_v1, import tasks, and other non-serializer paths; treats
  isolated entries the same as deleted ones (referral silently set to None)
- On restore, isolated referrals are cleared to None so restore always succeeds


userlocalhost (approved):
How smart and important work 👍

PR #1693, opened 2026-04-23
feat(entity): improve autoname UX on entity edit page

hinashi:
close: https://github.com/dmm-com/airone/issues/3456
close: https://github.com/dmm-com/airone/issues/3457
close: https://github.com/dmm-com/airone/issues/3465

- Add validation to EntityAttrCreateSerializer and EntityAttrUpdateSerializer
to raise a ValidationError when name_order > 0 is set on an attribute whose
type is not in Entity.ITEM_NAME_SELECTABLE_TYPES (STRING / OBJECT / NUMBER).
Previously such configurations were silently ignored at runtime.
- Disable the autoname config button for attribute types that do not support
autoname (TEXT, BOOLEAN, DATE, GROUP, ROLE, and array variants)
- Show a live preview of the generated item name in the "アイテム名の登録方法"
row when ATTR mode is selected, computed from name_order / prefix / postfix

### DIFF
--- a/entity/api_v2/serializers.py
+++ b/entity/api_v2/serializers.py
@@ -295,6 +295,10 @@ class EntityAttrCreateSerializer(serializers.ModelSerializer):
                     "When specified object type, referral field is required"
                 )
 
+        name_order = attr.get("name_order", 0)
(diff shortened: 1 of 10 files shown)


userlocalhost (approved):
How great work it is!! Thank you very much for enhancing its feature 👍

PR #1694, opened 2026-04-25
fix(entry): record deleted_user on /api/v1/entry DELETE (#3460)

syucream:
The v1 delete_entry Celery task called entry.delete() without passing deleted_user, so Entry.deleted_user stayed None.


hinashi (approved):
LGTM!

PR #1695, opened 2026-04-26
fix(advanced-search): resolve attr type from entityAttrs for bulk update

syucream:
The bulk update will be broken on changing entity attributes because the updater guesses the schema from the first element  of the entries, without checking actual entity attr.


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.28.1",
+  "version": "0.28.2",
```
syucream, line 3: Release it plz 🙏

hinashi (approved):
LGTM!

PR #1696, opened 2026-04-26
fix(entry): preserve int type for NUMBER attributes (#3458)

syucream:
The number type stores values with float-typed. It implicitly convert float -> int on loading the values from DB.


hinashi (approved):
LGTM!

PR #1697, opened 2026-04-26
eslint: add path groups for external imports

syucream:
`npm run fix` breaks the file formats if we combine it with `npm run link:client`.  It mitigates the problem.


hinashi (approved):
LGTM!

PR #1699, opened 2026-05-02
feat(advanced-search): add retrieve_all option to search_entries

syucream:
Expand the max number of advanced search results, from 99999(hard-coded) to `MAXIMUM_RESULTS_NUM`


userlocalhost (approved):
Good catch! Thank you very much for making a smart feature 👍

hinashi (approved):
LGTM!

PR #1700, opened 2026-05-02
fix(entry): omit unset 'referrals' key from YAML export (#3459)

syucream:
If a user exports advanced search result without `参照アイテムも含める` option, omit the field.


userlocalhost (approved):
Reasonable fixing 👍 Thank you very much for its good work!

hinashi (approved):
LGTM!

PR #1701, opened 2026-05-06
fix(entry): clear input text after selecting in array_object autocomplete

syucream:
On the attribute value edit page, in fields are typed as array object, the autocomplete input field doesn't clear input values on a user select some value. Its a bit noisy, so the PR ensures to clear that.


userlocalhost (approved):
Good catch 👍 Thank you very much for fixing UX problem.

hinashi (approved):
LGTM!

PR #1708, opened 2026-05-11
Added an API to retrieve logined user infomation

userlocalhost:
This is for telling LLM to the current login user information.
The API path `/me` is following after the Slack.


hinashi (approved):
LGTM!

PR #1716, opened 2026-05-16
fix(entry): self-heal Attribute when all AttributeValues lost is_latest=True

syucream:
When concurrent imports race in Attribute.add_value (non-atomic create + add + unset_latest_flag), every AttributeValue can end up with is_latest=False. In that state, Attribute.is_updated short-circuits on last_value.value == recv_value and EntryUpdateSerializer.update skips add_value, so the broken invariant never recovers.

Force is_updated to return True when no AttributeValue has is_latest=True so the next write creates a fresh is_latest=True AV via add_value and unset_latest_flag restores the "exactly one is_latest=True" invariant.

### DIFF
--- a/entry/models.py
+++ b/entry/models.py
@@ -603,6 +603,14 @@ class Attribute(ACLBase):
             else:
                 return recv_value
 
+        # Self-heal: when the invariant "exactly one AttributeValue with
+        # is_latest=True" has been broken (e.g. by a race between concurrent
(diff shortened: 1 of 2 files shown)


userlocalhost (approved):
Thank you very much for making a robust function for keeping consistency of what the latest value is 👍

hinashi (approved):
LGTM!

PR #1717, opened 2026-05-16
fix: enable disallow_untyped_calls in mypy

syucream:
Enable disallow_untyped_calls globally and annotate the 17 untyped callees that were previously invoked from typed contexts

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -9,7 +9,7 @@ from django.conf import settings
 from airone.lib.log import Logger
 
 
-def main():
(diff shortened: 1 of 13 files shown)


userlocalhost (approved):
Good 👍 This kind of adding hints is very important not only for increasing maintainability but also getting a benefit of support from LLM, I guess. Thank you very much!

hinashi (approved):
LGTM!

PR #1718, opened 2026-05-16
use database replica for advanced search

syucream:
Currently POST method handlers forward DB requests to the master instance by default. The advanced search accepts POST requests but it doesn't store any data into the DB, so I want an option to refer replica instances in that case.


userlocalhost (approved):
How great feature it is!! It's a revolutionary function. Thank you very very much!! 💯

hinashi (approved):
LGTM!

PR #1721, opened 2026-05-23
Fix stray ^M control character in CSV export of array attributes

syucream:
Eliminate noisy ^M control character's on exported CSV has an array string typed values.


userlocalhost (approved):
Great!! That is the one what we want, recently! Thank you very much 👍

hinashi (approved):
LGTM!

PR #1723, opened 2026-05-23
fix: enable warn_unused_configs and extra_checks in mypy

syucream:
Both flags produce zero new errors on the current codebase:
- warn_unused_configs: detects dead [[tool.mypy.overrides]] sections, keeping the config honest as modules are added/removed.
- extra_checks: enables additional correctness checks (TypedDict item access, Final, stricter cast handling) that mypy leaves out of the default for backward compatibility.


userlocalhost (approved):
Thank you for utilizing mypy features more 👍

hinashi (approved):
LGTM!

PR #1725, opened 2026-05-25
Set high-lighting to triggered attribute and enable to check it easily.

userlocalhost:
Trigger configuration may make user confused because attribute value might be changed unexpectedly because of its action. This feature of this commit let user know which attribute is targeted by trigger configuration. By this feature, user can know what happened at the time attribute value was changed unexpectedly.

---
スクリーンショット 2026-05-25 16 24 11

## Execution Result
https://github.com/user-attachments/assets/ea3f5c9b-7ad4-4949-92a4-16de1cd82a85


hinashi (approved):
LGTM!

PR #1729, opened 2026-05-30
Remove unused type:ignore[misc] from example plugins

syucream:
mypy flags these decorator-level `# type: ignore[misc]` comments as unused (warn_unused_ignores) now that pagoda-plugin-sdk ships py.typed and the hook/ override decorators resolve as typed. Removes 10 in hello-world plugin and 4 in cross-entity plugin. No behavior change.


hinashi (approved):
LGTM!

PR #1730, opened 2026-06-01
Added new API to rollback item attribute values at specified datetime

userlocalhost:
The `/entry/api/v2/{attribute-value-id}/attrv_restore/` API can only retrieve just single attribute value.
But there is an use-case to rollback item(s) to be the one(s) that have values at specified date time.
New API `/entry/api/v2/rollback` enables it.


hinashi (approved):
LGTM!

PR #1731, opened 2026-06-01
Introduced new parameter "to" that describes termination date-time of getting activity records.

userlocalhost:
User may want to know what changes are from point-A to point-B. The point-A is current date-time by default, and user can modify it by `since` parameter. But there was no way to specify point-B (termination date time) except for `within_minutes` parameter. This PR introduced it by `to` parameter to specify point-B by date-time.


hinashi (approved):
LGTM!

PR #1733, opened 2026-06-06
Enhance typing more

syucream:
SSIA :)


userlocalhost (approved):
Great! It's important not only increasing maintainability, but also especially for getting good support from LLM recently I think.

hinashi (approved):
LGTM!

PR #1734, opened 2026-06-06
Enable to sort advanced search results

syucream:
Sort by string/text/object/group/role/date/datetime attrs and entry name from column headers.Number, boolean, array_* and named_object variants are deliberately out of scope (would require ES re-indexing) and reject with 400.


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.28.2",
+  "version": "0.28.3",
```
syucream: Release the new version please 🙇

hinashi (approved):
LGTM!

PR #1736, opened 2026-06-13
Allow admin's to show all the jobs created by another users

syucream:
Currently the job list is strictly limited by user scope, but its inconvenient to resolve troubles. It allows admin's to confirm all the jobs.

image


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.28.4",
+  "version": "0.28.5",
```
syucream, line 3: Check it please 🙏

hinashi (approved):
LGTM!

PR #1738, opened 2026-06-15
Fix entry details page 404 when no triggers exist

tsunoda-takahiro:
This PR fixes an issue where the entry details page could fail with a 404 error when no trigger records existed.

Previously, the trigger list API returned 404 when the queryset was empty. The entry details page always fetched trigger data, so that response was propagated to the UI as a not found error.

Changes:

Return an empty paginated result from the trigger list API when no triggers exist
Add a regression test for the empty trigger list case
Impact:

Entry details pages no longer fail just because no triggers are configured
Trigger list behavior is now consistent with standard list API expectations

### DIFF
--- a/trigger/api_v2/views.py
+++ b/trigger/api_v2/views.py
@@ -1,7 +1,6 @@
 from typing import Any
 
 from django.db.models import QuerySet
-from django.http import Http404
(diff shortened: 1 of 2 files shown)


hinashi (approved):
LGTM!

PR #1746, opened 2026-06-18
Added implementation to prevent changing values that are updated by others

userlocalhost:
The implementation of the /entry/api/v2/rollback API has been modified as follows:

For attribute values ​​targeted for rollback, if there are changes made by someone other than the user between the AttributeValue to be rolled back and the current AttributeValue, the attribute will be excluded from the rollback (the value will remain unchanged).
( The variable `v_before` is the target AttributeValue to be rolled-back. )


hinashi (approved):
LGTM!

PR #1753, opened 2026-06-28
Support select/multiselect types

syucream:
Allow to set values from predefined options. Its like select-option in HTML, select/multiselect types in Notion, etc.

image
image


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.28.5",
+  "version": "0.29.0",
```
syucream, line 3: review this plz at first 🙏

hinashi (approved):
LGTM!

PR #1754, opened 2026-06-28
Modernize Python code with StrEnum, dataclass slots, and f-strings

syucream:
Use modern python features


userlocalhost (approved):
Good! Thank you very much for modernizing its implementation 👍

hinashi (approved):
LGTM!

PR #1755, opened 2026-06-29
Enable to show attribute value of ARRAY_NAMED_OBJECT_BOOLEAN type

userlocalhost:
## CAUSION
This change is not compatible with current Elasticsearch index. You have to run following command to update its index dynamically.
```Python
$ uv run manage.py shell

from airone.lib.elasticsearch import ESS
es = ESS()
es.indices.put_mapping(
    index=es._index,
    body={"properties": {"attr": {"type": "nested", "properties": {
        "boolean": {"type": "boolean", "index": "true"}
    }}}},
)
```

## Summary
There are some ARRAY_NAMED_OBJECT_BOOLEAN typed Attributes that are special Attribute in some use-case. But there is no way to show its value.
This PR enables to show its value through React UI.

### Rest of works
* [x] adapting its type for Elasticsearch


hinashi (approved):
LGTM!

PR #1757, opened 2026-07-10
Added entry_form as a target for plugin-based customization

hinashi:
SSIA


userlocalhost (approved):
LGTM to enhance capability of plugin system.

PR #1758, opened 2026-07-10
Fixed results were missing when there were many search targets in SearchChainAPI

hinashi:
This fixes an issue where specifying multiple AND conditions for the `attrs` of a leaf node in `POST /api/v1/entry/search_chain` (and `POST /api/v2/entry/advanced_search_chain`, which uses the same serializer) would result in a 400 error—"Data overflow was happened. Please narrow down intermediate conditions"—even if the final result set was small.


userlocalhost (approved):
How great work it is !! Thank you very much for its great job 👍

PR #1760, opened 2026-07-11
Enable to display referred entry attribute value(s) on the entry details page

syucream:
It empowers the entry details to include related object attribute values without a page transition.


apiclient/typescript-fetch/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dmm-com/airone-apiclient-typescript-fetch",
-  "version": "0.29.0",
+  "version": "0.29.1",
```
syucream, line 3: check it plz 🙇

hinashi (approved):
LGTM!

PR #1765, opened 2026-07-15
Fixed update es document tools

hinashi:
SSIA


userlocalhost (approved):
Thank you very much for caring to run it at the production environment 👍

PR #1772, opened 2026-07-18
Enable coding AI agents to flexibly use multiple ways to run tests

syucream:
The full UT/IT's is so high cost; requires MySQL, ES connections, frequent resets. Its too heavy to use coding agents like GitHub Copiot, Claude Code, Codex etc. It allows another quick ways to do that, sqlite based, with keepdb.


userlocalhost (approved):
Thank you very much for improving test acceleration 👍

hinashi (approved):
LGTM!

PR #1774, opened 2026-07-22
Add setup-pagoda-env composite action and adopt it in CI

hinashi:
## Summary

Introduce a shared composite action for the CI environment setup that is
duplicated across pagoda and its downstream extension repositories, and adopt
it in pagoda's own workflows.

- **New**: `.github/actions/setup-pagoda-env` composite action. It installs uv
  and Python, installs the system dependencies (apt), optionally clones the
  Pagoda base, optionally places an extension into it, runs `uv sync`, and
  optionally prepares MySQL. Service containers and the actual checks stay in
  the calling workflow (composite actions cannot declare `services:`).
- **Python version is centralized** in the action (single source of truth), so
  callers no longer pass it and the matrix `python-version` is dropped.
- **`clone` input** lets pagoda's own CI reuse the action without cloning
  itself (`clone: false`, `base-dir: .`).
- **Adopted in pagoda's own workflows**: `build-core.yml` (lint/test) and
  `release-apiv2-client.yml` now reference the action via a local `./` path,
  replacing the duplicated uv/python/apt/uv-sync/mysql-prep blocks.

## Why

The same environment-setup block was duplicated across pagoda's own workflows
and every extension repo (twc, airone.custom.racktables, airone.react.custom,
airone-playwright). Centralizing it removes multi-repo drift (Python version,
action pins, apt packages, service setup).

## Downstream

Extension repositories will reference this action as
`dmm-com/pagoda/.github/actions/setup-pagoda-env@<sha>` (the org requires a
full-length commit SHA). Their PRs currently pin a fork commit for pre-merge
verification and will be repointed to the merged SHA once this lands.

## Notes

- `services:` cannot live in a composite action, so each caller keeps its own
  service-container block (identical across repos).
- No pip cache on setup-python: dependencies are installed via uv (its own
  cache), and `cache: pip` requires a dependency file in the workspace which
  fails in jobs that clone the base into a subdirectory.

### DIFF
--- a/.github/actions/setup-pagoda-env/action.yml
+++ b/.github/actions/setup-pagoda-env/action.yml
@@ -1,0 +1,114 @@
+name: "Setup Pagoda extension env"
+description: >-
+  Shared CI environment setup for Pagoda/AirOne extension repositories.
+  Installs uv and Python, installs the system dependencies, clones the Pagoda
+  base, optionally places the extension inside the base, runs `uv sync`, and
+  (optionally) prepares MySQL. Service containers and the actual checks stay in
+  the calling workflow (composite actions cannot declare `services:`).
+inputs:
+  airone-branch:
+    description: "Pagoda base branch to clone."
+    required: false
+    default: "master"
+  base-dir:
+    description: "Directory to clone the Pagoda base into (also the working directory for uv)."
+    required: false
+    default: "airone"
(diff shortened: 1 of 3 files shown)


tsunoda-takahiro (approved):
LGTM!

PR #1778, opened 2026-07-22
feat: add agent-driven N+1 detection workflow

syucream:
N+1 is a so tough problem in the most of web applications. pagoda based on Django is as well. This agent skill allows AI Agents to auto-detect, fix the some problems with counting similar queries (like `SELECT ... FROM ... WHERE id = 1`, `SELECT ... FROM ... WHERE id = 2`, ...).


userlocalhost (approved):
Wow! How sophisticated and ambitious feature to improvement software code by LLM by it-self, for me. Thank you very much for telling us this great attempts 👍

hinashi (approved):
LGTM!

PR #1779, opened 2026-07-23
Cache CI setup deps in setup-pagoda-env (uv, apt) and speed up uv install

hinashi:
## Summary

Reduce the per-job setup time of the `setup-pagoda-env` composite action,
which every Pagoda extension repo (custom_view, twc, ...) runs — often
across many matrix jobs. Three changes:

- **Cache uv downloads** — persist uv's global download/build cache
  (`~/.cache/uv`) with `actions/cache` so `uv sync` reuses wheels and built
  artifacts instead of re-downloading everything each run. The step runs
  after the Pagoda base is cloned so the lockfile exists and keys the cache
  precisely. The key is `(OS, Python, uv.lock hash)` only — intentionally
  branch-agnostic, since the lockfile hash fully determines the dependency
  set, so base branches resolving to the same lockfile share one cache.

- **Faster uv install** — install uv via `astral-sh/setup-uv` (prebuilt
  binary) instead of `pipx install uv`, which builds/installs through pip.

- **Cache apt archives** — cache the downloaded `.deb` files. apt is pointed
  at a runner-owned directory (`~/apt-cache`) via `Dir::Cache::archives`
  rather than the root-owned `/var/cache/apt/archives`, so `actions/cache`
  can read/write it without permission workarounds. On a hit, `apt-get
  install` reuses the `.debs` and skips the download. The package list is
  resolved once and hashed into the cache key.

## Notes

- Effects apply from the **second run onward** (first run only populates the
  caches).
- New `uses:` (`astral-sh/setup-uv`, `actions/cache`) are pinned to
  full-length commit SHAs per org policy.
- Follow-up: after merge, bump the pinned
  `setup-pagoda-env@<SHA>` reference in the extension repos (custom_view,
  twc) to the merged SHA so they pick up the caching.

### DIFF
--- a/.github/actions/setup-pagoda-env/action.yml
+++ b/.github/actions/setup-pagoda-env/action.yml
@@ -56,10 +56,10 @@ runs:
         # Python version is centralized here (base side) so every extension
         # repository stays in sync. Bump it in this one place.
         python-version: "3.14"
-        # No pip cache: dependencies are installed via uv (its own cache), and
-        # `cache: pip` requires a requirements/pyproject file in the workspace,
-        # which fails in jobs that clone the Pagoda base into a subdir without
-        # checking out a Python project at the workspace root.
+        # No pip cache: dependencies are installed via uv, whose download/build
+        # cache is persisted by the "Cache uv downloads" step below. `cache: pip`
+        # would also require a requirements/pyproject file at the workspace root,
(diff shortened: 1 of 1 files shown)


userlocalhost (approved):
Great!! Thank you very much for utilizing GitHub Actions's feature 👍

PR #1780, opened 2026-07-23
Fixed not to be able to update token expiration by readonly-user's self.

userlocalhost:
↓execution result
https://github.com/user-attachments/assets/51bdf8ce-a590-4a91-8dae-d96ee93cc243


hinashi (approved):
LGTM!

PR #1784, opened 2026-07-30
Fix advanced search export for (ARRAY_)NAMED_OBJECT_BOOLEAN attrs

userlocalhost:
_yaml_export_v2() dispatches array values by `atype ^ AttrType._ARRAY`, but the resulting NAMED_OBJECT_BOOLEAN (_NAMED|OBJECT|BOOLEAN) does not match `case AttrType.NAMED_OBJECT`, so it fell through to the fallback branch and aborted the export Job with an AssertionError. Handle both named-object types and export the boolean flag alongside the referral info so the value stays round-trippable (the importer ignores the extra key).


hinashi (approved):
LGTM!

PR #1785, opened 2026-07-30
Fix import not updating (ARRAY_)NAMED_OBJECT_BOOLEAN attribute values

userlocalhost:
Importing an item never changed the boolean flag of a named-object-boolean attribute, for two independent reasons:

- EntryImportEntitySerializer converted named-object values into {"name": .., "id": ..} and dropped the boolean flag, so the value handed to add_value() always carried the default False. Add a _named_object() helper that keeps the flag, accepting both shapes it can be exported in: nested in the referral ({name, entity, boolean}, yaml export) and beside the name key ({<name>: <entry name>, boolean}, AttributeValue.get_value()).

- Attribute.is_updated() had no branch for the non-array NAMED_OBJECT_BOOLEAN, so a change of the flag alone fell through the match and was reported as "not updated". Handle it next to NAMED_OBJECT and compare the flag. The array variant already compared it, so it is fixed by the serializer change alone.

### DIFF
--- a/entry/api_v2/serializers.py
+++ b/entry/api_v2/serializers.py
@@ -1288,6 +1288,37 @@ class EntryImportEntitySerializer(serializers.Serializer):
                         return ref_entry.id if ref_entry else 0
                 return None
 
+            def _named_object(
+                val: dict[str, Any],
+                refs: list[Entity],
+                is_boolean: bool,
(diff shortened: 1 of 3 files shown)


hinashi (approved):
LGTM!

PR #1786, opened 2026-08-01
feat: replace attribute reorder arrows with drag-and-drop

syucream:
Migrate the entity attribute list reordering UX from up/down arrow buttons to drag-and-drop using @dnd-kit. Each row now exposes a grab handle in a dedicated column; rows can be reordered by pointer drag or keyboard (Space to lift, arrows to move, Space to drop), preserving accessibility.


userlocalhost (approved):
Excellent!! Thank you very much for making its great feature 👍

hinashi (approved):
LGTM!

PR #1788, opened 2026-08-01
Support import preview

syucream:
The import feature just accepts input data, and overwrite existing values. So it'll sometimes occur an unexpected result. This import-preview enables users to confirm the changes imported values will trigger before applying it.


syucream:
Splitted this large PR to three parts:

1. https://github.com/dmm-com/pagoda/pull/1793
2. https://github.com/dmm-com/pagoda/pull/1794
3. https://github.com/dmm-com/pagoda/pull/1795

PR #1789, opened 2026-08-03
Improve the devcontainer-based local development environment

hinashi:
## Overview

Makes the devcontainer usable as a full local development environment. Until now a
fresh clone needed several manual workarounds before tests, DB initialization or
API client generation would run inside the container.

### Isolate build artifacts and caches (`6995c24a`)
- Mount `.venv`, `node_modules` and the `uv` / `npm` / `playwright` caches as named
  volumes, so Linux-only binaries are not shared with the host through the bind
  mount and the caches survive container rebuilds.
- Pre-create those mount points in the Dockerfile with the container user's
  ownership, because Docker seeds an empty named volume from the image and creates
  a root-owned directory when the path is missing.
- Install `default-jre` (needed by `openapi-generator-cli` to run its jar) and set
  up Claude Code (extension, `~/.claude` mount, `CLAUDE_CONFIG_DIR`) plus a
  `~/.gitconfig` mount for convenience.

### Pre-create the host paths that are bind-mounted (`fd7e4532`)
Docker creates a *directory* when the source of a bind mount is missing, so a
member without `~/.gitconfig` on the host ended up with a directory there and every
git command in the container failed with `unknown error occurred while reading the
configuration files`. A missing `~/.claude` was created as a root-owned directory
on Linux hosts, which the container user cannot write. Both are now created on the
host via `initializeCommand`, which runs before the container is created.

### Make `clear_and_initdb.sh` work against a fresh MySQL (`00a4dafb`)
A newly created MySQL container only has the root account, so the script failed
before it could initialize anything. It now creates and grants the database user
configured in the settings via the root account first, and drops the database with
`if exists` so the first run does not fail. The root password can be supplied
through `MYSQL_ROOT_PASSWORD`, and the docker path uses `docker exec -i` instead of
`-it` because the heredoc is piped in without a tty.

### Generate the API client without a docker daemon (`dede09d1`)
In the non-docker path, use the `@openapitools/openapi-generator-cli` npm wrapper
instead of `docker run`. It executes the same Java jar, so the generated code is
unchanged, but it also works where no docker daemon is reachable — such as inside
the devcontainer. The generator version is pinned to `7.24.0` by the new
`openapitools.json`. The `--docker` path keeps using the generator container image.

### Install the GitHub CLI (`d488d516`)
Add `gh` to the packages installed in the image, so pull requests and issues can be
handled from inside the container. It comes from the Debian repository, which is a
few versions behind but sufficient here.

### Persist `~/.config` (`d23da377`)
Tools that are authenticated inside the container write their credentials to
`~/.config`, so a rebuild made them lose the login — the GitHub CLI being the first
one affected. It is mounted as a named volume rather than from the host, because the
host configuration may belong to another platform.

### Restart local middleware automatically (`8dbed3b5`)
Add `restart: unless-stopped` to `mysql`, `elasticsearch` and `rabbitmq` so the
local environment comes back after a docker daemon or host restart. Also drop the
unused `name: rabbitmq1` label.

## Notes

- No production/application code is touched: only `.devcontainer/`,
  `docker-compose.yml`, `tools/` scripts and the new `openapitools.json`.
- The generated API client output is byte-identical; only the way the generator is
  invoked changed.


tsunoda-takahiro (approved):
LGTM！

PR #1793, opened 2026-08-05
Add an import preview for models (1/3: backend)

syucream:
## Overview

The import feature just accepts input data, and overwrite existing values. So it'll sometimes occur an unexpected result. This import-preview enables users to confirm the changes imported values will trigger before applying it.

## The stack

| | | review effort |
| --- | --- | --- |
| **#1793** ← you are here | preview foundation + model import preview (backend) | +937 / −28 |
| #1794 | the preview UI, applied to the model import | +1522 / −62 |
| #1795 | the same preview for items, + the guard against overwriting | +1098 / −15 |

Splits #1788. GitHub's stacked pull requests don't support forks, so all three
target `master` and each diff also contains the layers above it in this table
(#1795 shows +3553 in total). Read them top-down.

## This PR

The import overwrites whatever it finds, so the only way to learn what a file
does is to let it happen. This adds a read-only preview that reports what a
model import would change, before it changes anything.

- Extracts the row checks out of `import_instance()` (pure refactor, first commit)
- Adds the shared preview payload and `GET /job/api/v2/<id>/preview`
- Adds `POST /entity/api/v2/import/preview`

Two things worth knowing:

- **The preview never writes.** Running the real import in a transaction and
  rolling back would report the same thing, but it would take write locks on
  rows the user only asked to look at, and would leave the changes behind if
  the rollback ever failed to run.
- **It runs as a celery job**, because previewing costs as much as importing. Preview
  jobs are hidden from the job list but stay cancelable.

Nothing changes for existing callers — posting to the import endpoint still
imports straight away.


userlocalhost (approved):
LGTM. Thank you very much for making its feature.
It's a thoughtful method to use delayed processing using celery, I think 👍 

> The preview never writes

At first, I didn't understand why the `POST` method was used even tough the preview didn't actually perform a write operation, but I realized that it was because an `IMPORT_ENTITY_PREVIEW` job was being created internally.

hinashi (approved):
LGTM!

PR #1794, opened 2026-08-05
Show what a model import would change (2/3: UI)

syucream:
The UI for the model import preview added in #1793. The import dialog gains a
  second button — the import button still imports the moment it is pressed.

  - Waits for the preview job while reporting its progress, then shows what the
    file would create, update, leave alone, skip or fail on, with the old and new
    values side by side. Rows can be narrowed by action, paged, or taken as CSV.
  - Previewing a large file takes as long as importing it, so the wait is
    interruptible and closing the dialog stops the polling.
  - Adds a live e2e suite that drives this against a real server, since no mock
    can tell you whether a preview really writes nothing. It is not part of CI —
    see `e2e/live/README.md`.
  - Bumps the API client to 0.29.2, where the preview endpoints live.

2/3, splitting #1788: #1793 (merged) → **this** → #1795 (items).


userlocalhost (approved):
Good! Thank you very much for making FE processing. The idea of showing import preview by manipulating backend job but rendering it as real-time is cool 👍

hinashi (approved):
LGTM!

PR #1796, opened 2026-08-07
Introduce lite mode to execute coding AI agents in parallel

syucream:
Today, we usually execute multiple AI coding agents at the same time. We should provide git-worktree friendly, and lightweight way to verify the agent output without resource conflicts.

The lite mode doesn't require containers (memory-based, or file-based, e.g. SQLite, ES on memory, RabbitMQ on memory) lightweight testing (but I notice its not a perfect environment to test the exact behavior).

In my dev environment, it can accelerate the testing time with independency.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -191,3 +191,6 @@ helm/charts/*
 
 # hugo
 docs/.hugo_build.lock
+
+# lite dev mode state: per-checkout SQLite DB, search index, media, test logs
+.pagoda-lite/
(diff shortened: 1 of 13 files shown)


userlocalhost (approved):
Excellent invention 👍 I always admire your passion to save resource. And time is one of the most precious resource we have, doesn't it. Thank you very much for introducing lite mode to reduce test time by using lite-weight middle-wares and making code to manipulate them.

hinashi (approved):
LGTM!

PR #1800, opened 2026-08-12
fix: make entity/entry name search case-insensitive independent of co…

nakajima-dmm0:
### Summary
Changing the name column's collation from general_ci to ac_cs (case-sensitive) allowed registering entity/entry names that differ only by letter case, but as a side effect the default icontains search became collation-dependent, so searches no longer matched when the search term's case didn't match the stored name's case
Both EntityAPI (/entity/api/v2/?search=) and EntityEntryAPI (/entity/api/v2/<entity_id>/entries/?search=) now search against a Lower(name) annotation instead, keeping search case-insensitive regardless of collation
### Changes
entity/api_v2/views.py
Added NameCaseInsensitiveSearchFilter and pointed EntityAPI's search_fields at the name_lower annotation
Added Lower("name") annotation in EntityAPI.get_queryset
Updated AliasSearchFilter similarly to search against name_lower / aliases_name_lower
Added Lower("name") (and Lower("aliases__name") when with_alias is set) annotation in EntityEntryAPI.get_queryset

### MySQL Change Query
ALTER TABLE acl_aclbase MODIFY name VARCHAR(200) COLLATE utf8mb4_0900_as_cs;


hinashi (approved):
LGTM!

PR #1803, opened 2026-08-14
Make AdvancedSearchService user argument optional

hinashi:
## Summary
- change `AdvancedSearchService.search_entries` argument type from `User` to `User | None`
- change `AdvancedSearchService.apply_join_attrs` argument type from `User` to `User | None`
- align `search_entries` docstring for `user` with the actual optional behavior

## Why
- implementation already handles `user is None`
- this aligns type hints and docs with runtime behavior

## Validation
- `uv run mypy --install-types --non-interactive --config-file custom_view/pyproject.toml custom_view`
- result: `Success: no issues found in 217 source files`

### DIFF
--- a/entry/services.py
+++ b/entry/services.py
@@ -35,7 +35,7 @@ class AdvancedSearchService:
     @classmethod
     def search_entries(
         kls,
-        user: User,
+        user: User | None,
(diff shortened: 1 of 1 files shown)


userlocalhost (approved):

PR #1809, opened 2026-08-18
Fix: TypeError when boolean is passed to attrinfo.keyword in entry search API

nakajima-dmm0:
EntrySearchAPIRequest.validate_attrinfo_keyword_length in api_v1/entry/views.py called len(item["keyword"]) directly. When a client sent a JSON boolean (e.g. "keyword": true or "keyword": false) instead of a string, this raised TypeError: object of type 'bool' has no len() for true, and for false it bypassed the length check entirely but then failed downstream when AttrHint(keyword=False) was constructed, since pydantic v2 does not coerce bool to str, resulting in a generic "The type of parameter 'attrinfo' is incorrect" response.

The validator now converts a boolean keyword value to its string representation (True/False) before the length check, so both truthy and falsy boolean inputs are handled consistently and no longer raise an unhandled exception or an incorrect error response.

issue: https://github.com/dmm-com/airone/issues/3606

### DIFF
--- a/api_v1/entry/views.py
+++ b/api_v1/entry/views.py
@@ -48,6 +48,8 @@ class EntrySearchAPIRequest(BaseModel):
     @field_validator("attrinfo")
     def validate_attrinfo_keyword_length(cls, v: list[Any]) -> list[Any]:
         for item in v:
+            if isinstance(item["keyword"], bool):
+                item["keyword"] = str(item["keyword"])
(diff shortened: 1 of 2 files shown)


nakajima-dmm0:
https://github.com/dmm-com/pagoda/pull/1809/commits/f126f0bb6e6c4773a61301bb6d2d6ac29247faf7

The fix was reverted to handle situations such as empty keywords.

hinashi (approved):
LGTM!

PR #1811, opened 2026-08-18
delete click event

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3605

Fixed an issue where click events were removed and the content was not removed when navigating without modal access.


hinashi (approved):
LGTM!

PR #1812, opened 2026-08-20
Fix MySQL option spacing in clear_and_initdb

hinashi:
## Summary
- Replace full-width spaces between `--skip-ssl` and MySQL user options with normal spaces in `tools/clear_and_initdb.sh`
- Prevent `mysql: unknown option '--skip-ssl　-uroot'` during DB initialization

## Verification
- `bash -n tools/clear_and_initdb.sh`


userlocalhost (approved):

PR #1813, opened 2026-08-20
Add belonging groups and roles to the user retrieve API

tsunoda-takahiro:
## Summary

Backend part of showing the groups and roles a user belongs to on the user detail screen (`/ui/users/<id>`).

The frontend part follows in a separate PR, once the API client `0.29.3` has been published.

### Extracting `User.belonging_roles()` (refactoring)

The logic that aggregates the roles a user belongs to was inlined in `User.has_permission()`. It is now extracted into `User.belonging_roles()`.

A user can belong to a role through four paths:

| Field | related_name | Meaning |
|---|---|---|
| `Role.users` | `user.role` | member (assigned to the user directly) |
| `Role.groups` | `group.role` | member (through a group) |
| `Role.admin_users` | `user.admin_role` | administrator (assigned to the user directly) |
| `Role.admin_groups` | `group.admin_role` | administrator (through a group) |

`has_permission()` only needed the union of those four paths, but the screen needs to know *why* a role is granted. So the new method returns `BelongingRole` entries that keep the provenance (`is_direct` / `is_admin` / `via_groups`), aggregated into one entry per role. `has_permission()` now simply consumes it, and its results are unchanged.

### Reducing the number of queries

While moving the logic, resolving the roles of belonging groups was changed from two queries per group to a single pair of queries regardless of the number of groups. This matters because `has_permission()` is called repeatedly on a hot path (per entry).

A regression test asserts that the query count does not grow with the number of groups (`test_belonging_roles_issues_constant_number_of_queries`).

### New fields on the retrieve API

`UserRetrieveSerializer` (`GET /user/api/v2/<pk>/`) now returns:

- `groups`: `id` / `name` / `is_direct`
  - `is_direct=false` means the user belongs to it only through a hierarchical superior group, not directly
- `roles`: `id` / `name` / `is_direct` / `is_admin` / `via_groups`
  - `is_direct` — the user is registered to the role itself
  - `is_admin` — the user has administrative privilege through any of the paths
  - `via_groups` — `id` / `name` of the groups that grant the role

The provenance is returned because listing hierarchical groups and group-granted roles without distinction makes it impossible to tell from the screen where a permission comes from.

Visibility is unchanged: the existing `UserPermission` still restricts `retrieve` to superusers, the user themselves, and their parent user.

## API client version

`apiclient/typescript-fetch/package.json` is bumped `0.29.2` -> `0.29.3`. `release-apiv2-client.yml` publishes it once this is merged.

A follow-up PR will then bump the root `package.json` / `package-lock.json` together with the frontend implementation. Since this PR contains no frontend change, CI passes as-is.

## Tests

`user/tests/test_model.py`

- `test_belonging_roles` — all four paths are collected, hierarchical superior groups are traversed, inactive roles and roles of unrelated groups are excluded, and `is_direct_belonging=True` behaves as expected
- `test_belonging_roles_aggregates_multiple_paths` — a role reachable through several paths is reported once, with the provenance merged
- `test_belonging_roles_issues_constant_number_of_queries` — the query count does not grow with the number of groups

`user/tests/test_api_v2.py`

- `test_get_user_groups_and_roles` — response contents (inherited group, group-granted role, admin role)
- `test_get_user_groups_and_roles_when_user_belongs_to_nothing` — empty arrays when the user belongs to nothing
- `test_get_user_excludes_deleted_groups_and_roles` — logically deleted groups and roles are not returned

### Verified locally

- `uv run ruff check .` and `uv run ruff format --check .` pass
- `uv run mypy` — no errors under `user/`
- Whole backend suite: 1920 tests pass. The only failure is `custom_view_sample.tests.test` raising `ModuleNotFoundError`, a test module that is already broken on master and unrelated to this change
- No permission-check regressions across all apps, including `acl` / `role` / `group` / `entry` / `entity` / `custom_view`


hinashi (approved):
LGTM!

PR #1814, opened 2026-08-20
Show belonging groups and roles on the user detail screen

tsunoda-takahiro:
## Summary

Frontend part of showing the groups and roles a user belongs to on the user detail screen (`/ui/users/<id>`).

Follow-up to #1813, which added `groups` and `roles` to the user retrieve API. This PR bumps the API client to the published `0.29.3` and consumes those fields.

### API client bump

- `package.json` — `optionalDependencies` range `^0.29.1` -> `^0.29.3`
- `package-lock.json` — regenerated for `0.29.3`

The dependency stays in `optionalDependencies`. `npm install <pkg>@<ver>` moves it to `dependencies` on its own, which would change `npm ci` behaviour and break local development without the published package, so the range was edited in place and the lock regenerated with `npm install --package-lock-only`.

### Two new rows on `UserForm`

`ElemGroups` and `ElemRoles` are added right after the name row, so the information is visible without scrolling past the token settings.

- Each chip links to the group / role detail screen via `groupPath()` / `rolePath()`
- Directly assigned groups and roles are filled chips
- A group inherited from a hierarchical superior group is an outlined chip labelled `<name> (継承)`
- A role granted through a group is an outlined chip labelled `<name> (<group> 経由)`
- A role the user administrates is labelled `<name> (管理)`, combined as `<name> (管理 / <group> 経由)` when both apply
- A placeholder message is shown when the user belongs to nothing
- The rows are hidden while creating a user, where there is nothing to show yet

The rows are read only. Editing stays in `GroupEditPage` / `RoleEditPage`, so there is a single place that owns the membership.

Visibility is unchanged: this screen is already restricted to superusers, the user themselves, and their parent user by `UserPermission` on the backend.

### Test fixture update

`ChangeUserAuthModal.test.tsx` builds a `UserRetrieve` literal with a type annotation, so it needs the new required fields to compile against `0.29.3`. This is why the frontend change could not be included in #1813, where the client was still `0.29.1`.

## Tests

`UserForm.test.tsx`

- `should not show belonging groups and roles on creating a user`
- `should show belonging groups and roles` — chip labels and the link target of each chip, covering an inherited group, a group-granted role and an administrative role
- `should show placeholders when the user belongs to nothing`

The existing `should provide user editor` test is unchanged in behaviour; the duplicated render setup was factored into a `renderUserForm` helper.

### Verified locally

- `npx eslint frontend e2e` — 0 errors (33 pre-existing warnings)
- `npx biome format` — clean on the changed files
- `npx knip` — clean
- `npm run test` — 148 suites / 810 tests pass, coverage above every threshold (statements 76.92% / branches 56.47% / functions 60.35% / lines 76.28%)
- `webpack build --entry ./frontend/src/App.tsx --mode development` and `npm run build:lib` (including `tsc --project tsconfig.lib.json`) both succeed

### DIFF
--- a/frontend/src/components/user/ChangeUserAuthModal.test.tsx
+++ b/frontend/src/components/user/ChangeUserAuthModal.test.tsx
@@ -26,6 +26,8 @@ describe("ChangeUserAuthModal", () => {
       created: "",
     },
     authenticateType: UserRetrieveAuthenticateTypeEnum.AUTH_TYPE_LOCAL,
+    groups: [],
+    roles: [],
   };
 
   test("should render a component essentially", async () => {
--- a/frontend/src/components/user/UserForm.test.tsx
+++ b/frontend/src/components/user/UserForm.test.tsx
@@ -2,7 +2,10 @@
  * @jest-environment jsdom
  */
 
-import { UserRetrieveAuthenticateTypeEnum } from "@dmm-com/airone-apiclient-typescript-fetch";
+import {
+  UserRetrieve,
+  UserRetrieveAuthenticateTypeEnum,
+} from "@dmm-com/airone-apiclient-typescript-fetch";
 import { zodResolver } from "@hookform/resolvers/zod/dist/zod";
 import { render, renderHook, screen } from "@testing-library/react";
 import { useForm } from "react-hook-form";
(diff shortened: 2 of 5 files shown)


userlocalhost (approved):
Good work 👍 Thank you very much for this important work.

hinashi (approved):
LGTM!

PR #1819, opened 2026-08-24
delete v1 rerun

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3611
## Summary

Remove the job rerun API tests and related rerun coverage from the test suite.

## Background

This change is part of the effort to remove the rerun feature from the UI.  
The rerun endpoint is not used for initial job execution, so the rerun-specific test coverage can be removed without affecting normal create/edit flows.

## Changes

- Removed rerun-related tests from `api_v1/tests/job/test_api.py`
- Removed rerun-related tests from `job/tests/test_view.py`
- Cleaned up unused imports and leftover test code after test removal
- Verified Python lint passes with `uv run ruff check .`

### DIFF
--- a/api_v1/job/urls.py
+++ b/api_v1/job/urls.py
@@ -4,6 +4,5 @@ from . import views
 
 urlpatterns = [
     re_path(r"^$", views.JobAPI.as_view()),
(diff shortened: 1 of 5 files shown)


hinashi (approved):
LGTM!

PR #1822, opened 2026-08-28
Support MySQL clients without skip-ssl

hinashi:
## Summary
- Detect the SSL-disabling option supported by the local MySQL client during DB initialization
- Use `--skip-ssl` when available and fall back to `--ssl-mode=DISABLED` for clients that expose `--ssl-mode`

## Verification
- `bash -n /workspace/airone/tools/clear_and_initdb.sh`

### DIFF
--- a/tools/clear_and_initdb.sh
+++ b/tools/clear_and_initdb.sh
@@ -35,6 +35,18 @@ parse_argv() {
   done
 }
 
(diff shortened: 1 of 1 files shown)


tsunoda-takahiro (approved):
LGTM!

PR #1825, opened 2026-08-29
fix: allow clearing EntityAttr default values and stop re-filling them

syucream:
The default value couldn't be cleared, thats a problem so it resolves.


userlocalhost (approved):
Good 👍 Thank you very much for fixing its problem.

PR #1826, opened 2026-08-31
Add local Hugo release script

hinashi:
## Summary
- Remove the GitHub Actions Hugo release workflow that no longer works under GitHub IP restrictions.
- Add `tools/release_hugo.sh` so maintainers can build Hugo docs locally and publish them to `gh-pages`.
- Clean the temporary `gh-pages` worktree before copying `docs/public`, so stale generated files are removed on release.

## Verification
- `bash -n tools/release_hugo.sh`
- `git diff --check -- tools/release_hugo.sh .github/workflows/release-hugo.yml`


userlocalhost (approved):
Great! Thank you very much for making this utility tools 👍

PR #1827, opened 2026-08-31
Allow mypy ratchet CLI output in Ruff

hinashi:
## Summary
- Allow `tools/mypy_ratchet.py` to use `print` as a CLI reporting script
- Update the Ruff per-file ignore comment to include `mypy_ratchet.py`

## Verification
- `uv run ruff check --output-format=github tools/mypy_ratchet.py`
- Reproduced the `build-core.yml` staticchecks sequence in a clean temporary worktree:
  - `uv run python tools/mypy_ratchet.py --base-ref 0080eadcd8b50a89f341e0a6ea898da81b2f65fa`
  - `uv run ruff check --output-format=github .`
  - `uv run ruff format --check .`


tsunoda-takahiro (approved):
LGTM!

PR #1828, opened 2026-08-31
Adding an unprocessed mark

nakajima-dmm0:
## Overview

Added a reload indicator to the Entry detail page while an Entry update is in progress, consistent with the existing Entity behavior.

## Changes

- Added `has_ongoing_changes` to the Entry API response
- Determine the update state from `STATUS_CREATING` and `STATUS_EDITING`
- Set `STATUS_EDITING` before dispatching the asynchronous update job
- Connected the update state to the Entry detail page header
- Added `hasOngoingChanges` to the generated `EntryRetrieve` TypeScript type

## Verification

- The reload indicator is displayed while the update job is pending, including when the worker is stopped
- The indicator disappears after the update is completed
- Confirmed that `npm run typecheck` passes


hinashi (approved):
LGTM!

PR #1829, opened 2026-09-01
add unprocessed mark

nakajima-dmm0:
## Overview

Updated the frontend to display a reload indicator when an Entry has ongoing changes.

## Changes

- Added support for `has_ongoing_changes` in the Entry API response
- Displayed the existing reload icon on the Entry detail page when an update is in progress
- Connected `hasOngoingChanges` from `EntryRetrieve` to `PageHeader`
- Updated the generated TypeScript API client type definitions

## Behavior

The reload icon is displayed while an Entry update job is pending or being processed, including when the worker is stopped. The icon disappears after the update is completed.


hinashi (approved):
LGTM!

PR #1830, opened 2026-09-01
Added a set-type method to prevent duplicates

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3630
## Overview

Prevented duplicate attribute names from being displayed in the Advanced Search Join modal.

## Changes

- Removed duplicate attribute names from the autocomplete options
- Removed duplicate attribute names from the initially selected values
- Ensured each attribute name is displayed only once

## Verification

- Confirmed that duplicate attribute names are no longer shown in the autocomplete list
- Confirmed that duplicate selected chips are removed when the modal is opened


hinashi (approved):
LGTM!

PR #1834, opened 2026-09-02
chore(deps): sync djangorestframework version

hinashi:
## Summary
- Sync root and plugin SDK djangorestframework pins with the 3.17.2 upgrade from #1832
- Regenerate the root and plugin SDK uv.lock files

## Verification
- `uv --directory /workspace/airone sync`


tsunoda-takahiro (approved):
LGTM!

PR #1835, opened 2026-09-02
fix pagenation

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3631

## Summary

Fix advanced search pagination when join filters are applied.

## Changes

- Continue fetching Elasticsearch result pages until 100 records remain after join filtering.
- Preserve the existing SQL/Python join implementation.
- Return the correct result count when fewer than 100 records match.
- Show a loading indicator while join-filtered results are being processed.
- Add a regression test covering cases where the first page does not contain enough matching records.

## Notes

No Elasticsearch query, schema, or document structure changes are included.


hinashi (approved):
LGTM!

PR #1836, opened 2026-09-02
Modify display conditions

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3624

Filter the item change history to display only records where the item name has changed.

History records are preserved in the database. Records created by attribute updates or other changes that do not modify the item name are excluded only from the UI display.


hinashi (approved):
LGTM!

PR #1842, opened 2026-09-04
Prevent Webpack from emitting extra UI chunks

hinashi:
## Summary

Prevent the frontend Webpack build from emitting many additional `*.ui.js` and vendor chunk files under `static/js`.

## Changes

- Disable automatic chunk splitting for the frontend bundle.
- Limit the generated bundle to a single chunk, `static/js/ui.js`.

## Validation

- Webpack emitted only `static/js/ui.js` in the dedicated worktree.
- The build reports four pre-existing TypeScript errors in unrelated frontend files and exits with code 1.


userlocalhost (approved):
LGTM 👍

PR #1844, opened 2026-09-07
preserve item name filter when adding attribute join

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3634

### Summary
Preserve the existing item name filter when adding an attribute join in Advanced Search.

### Changes
- Keep the current hint_entry search condition when updating join attributes.
- Added a regression test to verify that the item name filter is preserved.


hinashi (approved):
LGTM!

PR #1845, opened 2026-09-07
fix case and with insenitive entry search

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3641

Update entry search to support case-insensitive and width-insensitive matching for entry names and aliases.


hinashi (approved):
LGTM!

PR #1847, opened 2026-09-08
remove duplicate calls

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3648

Fix duplicate API requests when paginating Advanced Search results.

The `entry/api/v2/advanced_search/` API was called twice when navigating to page 2 or later because the page number was managed separately in the URL and React state. Centralize page management in the URL so that each page change triggers only one API request.

Add a regression test to verify that navigating to the second page results in exactly one additional API call.


hinashi (approved):
LGTM!

PR #1850, opened 2026-09-09
Implementation of search that handles full-width and half-width characters

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3660

Add case- and width-insensitive filtering for user search candidates in group and role management screens, including full-width and half-width Katakana support.


hinashi (approved):
LGTM!

PR #1852, opened 2026-09-09
Changed the offset limit to the maximum number of items.

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3654

Fix advanced search result count display so the displayed count never exceeds the total number of search results when loading paginated joined-search results.


hinashi (approved):
LGTM!

PR #1856, opened 2026-09-10
Adding editing permissions for parent users and eliminating screen tr…

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3667

Allow a parent user to update the Access Token lifetime of a read-only user they created.

Keep the user on the edit page after saving so the updated expiration time can be reviewed. Ensure the form can be edited and saved again after the first update, and add frontend and Python API regression tests.


hinashi:
@nakajima-dmm0 
新規ユーザ作成時はページ遷移するようにして欲しいです。
image

hinashi (approved):
LGTM!

PR #1858, opened 2026-09-11
Reduce report-coverage job parallelism in build-core.yml

hinashi:
## Summary

- 8419496506ae951c2d9d2a293f6f131b97e06e05 separated octocov into a `report-coverage` job for minimal permissions, but that job kept the same `matrix.target` as `test`, so it now runs one extra job per target (test N + report-coverage N instead of N).
- Collapse `report-coverage` into a single job. It downloads all per-target `coverage-*` artifacts into `coverage-reports/<target>/coverage.xml`, and `.octocov.yml`'s `coverage.paths` now uses a glob (`coverage-reports/**/coverage.xml`) so octocov merges them, matching the single-job shape `report-coverage` already has in build-frontend.yml.
- The permission-minimization intent from the original change (`test` no longer holds `pull-requests: write`) is unchanged.

## Test plan

- [ ] CI (build-core.yml) passes on this PR
- [ ] `report-coverage` runs as a single job and posts exactly one coverage comment on the PR
- [ ] The coverage numbers in that comment are unchanged from before this PR

### DIFF
--- a/.github/workflows/build-core.yml
+++ b/.github/workflows/build-core.yml
@@ -145,20 +145,17 @@ jobs:
 
   report-coverage:
     runs-on: ubuntu-24.04
-    needs: [setup_python, test]
+    needs: test
     permissions:
       contents: read
       pull-requests: write
-    strategy:
-      fail-fast: false
-      matrix:
-        target: ${{ fromJson(needs.setup_python.outputs.targets) }}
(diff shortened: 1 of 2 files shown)


tsunoda-takahiro (approved):
LGTM!

PR #1859, opened 2026-09-11
Adding read-only users to the user management screen: Frontend implementation

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3666

Update the user management frontend to keep the user edit page open after updating an existing user, while redirecting to the user list after creating a new user.

Also prevent the unsaved-changes confirmation dialog from appearing during successful user creation and allow users to edit and save multiple times without leaving the page.


hinashi (approved):
LGTM!

PR #1861, opened 2026-09-12
Build ES queries with typing

syucream:
Current ES queries are just dict values. Its so unsafe, we're sometimes confused with that. Recent ES client module provides types so this PR uses it, and combines our own types to mitigate the type safety issues.


userlocalhost (approved):
Great work! This must help future coding work 👍 Thank you very much for your great improvement.

hinashi (approved):
LGTM!

PR #1862, opened 2026-09-14
Permission management for items linked to attributes

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3642

Implemented backend-only access control for referral item suggestions.
The API now filters out items that the current user cannot view and returns a has_restricted_items flag when any candidates are hidden due to permissions.


hinashi (approved):
LGTM!

PR #1864, opened 2026-09-14
Fixed an issue where filter() was being called while the attribute array was uninitialized

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3676

Handle undefined attribute arrays before calling `filter()` in the auto-name preview.


hinashi (approved):
LGTM!

PR #1865, opened 2026-09-15
Support for half-width and full-width uppercase characters

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3663

Add case- and width-insensitive search to candidate selectors across item, group, role, category, model, and trigger management screens, with regression tests.


hinashi (approved):
LGTM!

PR #1868, opened 2026-09-16
Fix/advanced search autocomplete input

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3669

Fix advanced search group and role autocomplete input

Enabled text input and fuzzy matching for `array_group` and `array_role` bulk-update fields. Users can now replace existing selections, search by keyword, and select matching group or role candidates. Added and updated tests to verify input handling and API search behavior.


hinashi (approved):
LGTM!

PR #1869, opened 2026-09-17
Fixed a bug that may conflict cache of Entity and EntityAttr

userlocalhost:
## Summary
Previous implementation use history_id as cache-key of cache-table. But, each cache context of Entity and EntityAttr are indipendent. So it's necessary to separate name spaces of them. This commit fixed it.

## In detail
The `history_cache_key()` method is a key point of this PR that separate name-space by each cache-context. This method is used whenever cache is saved/fetched, then each cache context is stored to separated(proper) name spaces.


hinashi (approved):
LGTM!

PR #1870, opened 2026-09-17
Type specification correction

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3642

Fix the OpenAPI response schema for entry attribute referrals to correctly represent the object response and prevent the generated TypeScript client from treating it as an array.


hinashi (approved):
LGTM!

PR #1872, opened 2026-09-18
Modified to reflect the filtered results of reference items in the file output frontend side

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/3664

Fix advanced search CSV/YML exports to respect referral item filters, including referral name and included/excluded referral models.


hinashi (approved):
LGTM!

PR #1874, opened 2026-09-18
Detailed error log

nakajima-dmm0:
issue: https://github.com/dmm-com/airone/issues/1990

## Description

Limit job log display to 300 characters and append `...` when the log exceeds the limit.

For Entry import jobs:

- Display the number of successfully imported entries separately from the requested count.
- Store detailed error messages for up to five failed entries.
- Exclude the previous aggregate error summary from the job log.
- Preserve the warning status when partial failures occur.
- Extract and display the actual validation error messages only.

Added and updated tests for import result counts, error messages, five-error limit, and job log truncation.

image


hinashi (approved):
LGTM!

